Issue #50, opened 2020-05-04
All distributions need an issued date/time


ajtucker:
When fetching the latest distribution for a dataset, the list of distributions is sorted by issued date.

PR #51, opened 2020-05-12
Workaround for #50

ajtucker:
Implement workaround for distributions with no issued dates.


cadmiumcat:
Is this PR still relevant?

mikeAdamss:
yes, its an enhancement to avoid us using a long standing work around, but the work around works so it's relevant but not a priority.

cadmiumcat:
So this PR needs some context added so that someone can review it.

Otherwise, it can be changed for a draft request as it's not ready for review

RedWalters:
Looks good, makes sense
All tests pass on my messy local as well

RedWalters (approved):

PR #80, opened 2020-09-24
NHS Digital changes

ajtucker:
It turns out that there was only a minor change to the NHS Digital page (Publication date becomes Publication Date). I've made this more robust.

### DIFF
--- a/gssutils/metadata/base.py
+++ b/gssutils/metadata/base.py
@@ -48,7 +48,7 @@ class Metadata:
         elif name in self._core_properties:
             super().__setattr__(name, value)
         else:
-            raise AttributeError(f'Unkown attribute {name}')
+            raise AttributeError(f'Unknown attribute {name}')
 
     def get_unset(self):
         for local_name, profile in self._properties_metadata.items():
--- a/gssutils/scrapers/nhs_digital.py
+++ b/gssutils/scrapers/nhs_digital.py
@@ -31,24 +31,30 @@ def scrape(scraper, tree):
             dataset.landingPage = urljoin(scraper.uri, href)
             article_tree = html.fromstring(scraper.session.get(dataset.landingPage).text)
             article_type = article_tree.xpath("//span[contains(concat(' ', @class, ' '), ' article-header__label ')]/text()")[0]
-            assert article_type == 'Publication', 'Expecting publication'
+            assert article_type.startswith('Publication'), 'Expecting publication'
             details_node = article_tree.xpath("//dl[@class='detail-list']")[0]
             details = {}
             for node in details_node:
                 if node.tag == 'dt' and node.get('class') == 'detail-list__key':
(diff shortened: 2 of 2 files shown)


cadmiumcat (dismissed):
Thanks for starting on the PRs 🥰 

I think this a good chance to add some comments on what each section is doing, and adding some white-space between them to make it more readable. 

Should we raise some issues from the 2 to do tasks?

josepajay:
It looks good to me

ajtucker:
Please raise these change requests more generally and we can discuss the policies we want for spacing and comments, rather than block fixes.

What we need for this PR is to double-check the tests pass against the current NHS Digital pages and that the tests are still sensible.

cadmiumcat:
Thanks Alex. 

I didn't intend to block you - I actually find it quite hard to understand what's going on, so I thought it might be a good chance to improve those things. It would also help more junior devs contribute to the PR review.

I have not tried the code so I can't confirm that it works - I am also not sure how to check that it works. If this is time sensitive, then perhaps @mikeAdamss can do it and approve it. 

Otherwise, I believe you can override the review block and merge it.

ajtucker:
Thanks @cadmiumcat ; I didn't mean to be short!

I completely agree with the need for explanations and clarity, though as that's a more general issue I think we should separate it out from this specific issue so we can move on. Feel free to add some issues to the backlog.

For testing, we need to figure out how the fixtures, specifically `scrape.yml`, is going to work alongside using PRs and merging into main.

I propose that a reviewer of the PR needs to run any tests that check the functionality that has been fixed/updated before merging and then updating the `scrape.yml` file in the main branch.

For instance, in this branch, we could run:
```
pipenv run behave -D record_mode=all --tags=-skip -n 'NHS' features/scrape.feature
```

By passing `-D record_mode=all`, the tests will use the live/current web pages, rather than replaying what is in `scrape.yml`. They will record any new HTTP requests/responses to `scrape.yml`, but rather than add these changes to the branch, I think we should first merge the issue branch, run the tests again (with `record_mode=all`), then update `scrape.yml` on the main branch.

This should help keep the issue branches mergeable.

cadmiumcat:
We might have to agree to disagree on that one. I think:
✔️ Making small changes that add context and improve redability is part of _every PR_ 
✔️ It saves everyone time in the long run 
✔️ The insurmountable task of tidying everyone else's code becomes less horrifying

📞 The issue of the ymls does need addressing (I hadn't realised how they work) - perhaps we should have a call about it and loop everyone else in. 

Results from running tests (-n 'NHS'): 
```
2 features passed, 0 failed, 7 skipped
3 scenarios passed, 0 failed, 67 skipped
14 steps passed, 0 failed, 313 skipped, 0 undefined
Took 0m26.818s
```

cadmiumcat (approved):

PR #96, opened 2020-10-08
Issue 82 - Raise appropriate error when page type not supported by scraper

josepajay:


gssutils/scrapers/ons.py
```diff
@@ -13,7 +13,7 @@
 ONS_DOWNLOAD_PREFIX = ONS_PREFIX+"/file?uri="
 
 
-def scrape(scraper, tree):
+def scrape(scraper, tree, page_type_check = True):
```
mikeAdamss, line 16: I see where you're coming from but this won't work sadly, this `scrape` function is one of about a dozen that can be called from this list: https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/scrapers/__init__.py , (this one is `ons.scrape`). 

It means we'd struggle to pass these sorts of arguments in as each 'scrape' function gets automatically called based on matching the url, so we need to try and keep their function signatures the same.

One possible suggestion is use an environment variable, something fairly explicit like `DISABLE_ONS_SCRAPE_TYPE` (I'm bad at naming, but you get the idea).

Then you can do something like:
```
if not os.environ.get('DISABLE_ONS_SCRAPE_TYPE'):
       # do the type check
```

Doesn't have to be that specifically if you'd rather go another way, but we can't change the signature.

If you do go with an env var can you add a note to the readme as well please.

gimme a shout if I'm explaining that badly 👍

mikeAdamss (commented):
Code is great and test pass 👍 but we need to take a different approach for turning off the type check (my bad, should have detailed that properly). Have left some explanation.

gssutils/scrapers/ons.py
ajtucker: Agreed that it's tricky to use a different function signature for specific scrapers.

I wonder whether this is a bit of feature creep that we need to factor out into a separate issue?

What is the extra feature we're after? For automation, there's a balance between coping with strange results but within some bounds where we're pretty sure the automation will still do the right thing, and coping with unexpected results where we need to fail fast / bail with a reasonable error message. I think that while we're in development mode, it makes sense to be less strict about these strange results so that we can work out what's going on, whereas when the code is up and running in an automated pipeline, we need to be more strict about raising exceptions.

I think there's an argument for being able to turn off strict checking while we're developing transforms, but think we should create a separate issue/feature so it can be considered across all scrapers. It doesn't have to be a big feature!

mikeAdamss: Makes sense, lets un-scope creep this one then :smile:

josepajay: Ok, I have reverted the method signature and removed the type check based on a variable

mikeAdamss (approved):
looks good :smiley:

PR #98, opened 2020-10-12
Revert "Issue 90"

cadmiumcat:
Reverts GSS-Cogs/gss-utils#91

I agree with @ajtucker that there are issues when running all test (`pipenv run behave -D record_mode=none --tags=-skip`)


cadmiumcat:
@mikeAdamss is making changes in #99 instead of reverting

PR #99, opened 2020-10-12
Fixes/issues 82 90

mikeAdamss:
Fixes broken tests from recent merges, various:

* Add "static_adhoc" to ONS scraper allowed types to support a legacy test.
* Update vcr fixture for missing url.
* Add a guard for legacy tests calling seed where seed is none.
* Added a families to original seed test (as "families" wasn't a thing back then).

In theory that's everything back to passing, though at least one other person should definitely run the whole suite to make sure.

Run the tests :within a locally cloned gssutils repo
```
pipenv sync
pipenv install -r test-requirements.txt
pipenv run behave -D record_mode=none --tags=-skip
```

PR #100, opened 2020-10-12
Issue 76

mikeAdamss:
Create CSVWMapping() needs to work with csv.gz files as well as csv.

Note - probably worth running all test against this branch this after the fixes here: https://github.com/GSS-Cogs/gss-utils/pull/99 go in (and before merging this), as it's currently near impossible to know if this is breaking anything else.

You can check this specific functionality via:
```
pipenv sync
pipenv install -r test-requirements.txt
pipenv run behave -D record_mode=none -n "CSVWMapping with compressed CSV" --tags=-skip
```


ajtucker:
One thing we need to think about is that the CSV-W JSON file includes a link to the CSV file as part of the CSV-W discovery requirement. I think it should be pointing back to the `.csv` file rather than the `.csv.gz` file, even if the former doesn't exist.

It's all a bit of a workaround TBH and maybe a better approach would be to transparently compress stuff at the filesystem level if the concern is storage space. Network bandwidth is already considered as files will be automatically gzipped as required when transmitted by HTTP.

ajtucker (commented):
Looks good. Thanks for adding the check for the file ending. We could go further and add tests that run csv2rdf and look at the output, but that's probably over the top for this change.

Happy to merge.

ajtucker (approved):
Happy to merge.

PR #103, opened 2020-10-13
Fix for test automation

mikeAdamss:
combines fix to 82 and 90 with automation fix from 101. Also adds a cubes tweak to make it all work without assumptions about where the code is being ran.


ajtucker (approved):
Looks good.

Liking Swiss Family Robinson.

PR #110, opened 2020-10-28
Scraper should now determine whether the page is on the COVID19 page …

RedWalters:
…or a regular stats page and run relevant scraper. Feature also added to test the COVID19 page. Expect there to be issues with various landing pages since there seems to be many different layouts.

Comments and best practices are still needed. 

Support for various additional landing pages will need to be added - currently the 5 covid19 datasets managed to have 3 different layout landing pages, these _should_ be working correctly. Any issues please raise with me and I'll fix them


mikeAdamss:
sent JJ some info on the behave stuff (we need to document that really).

mikeAdamss (approved):
works

PR #112, opened 2020-11-16
vcr fixture helper script

mikeAdamss:
Helper script plus documentation for untangling things where we have issues with vcr request/response recordings.

### To review:

Have a play locally, should make sense.


ajtucker (approved):
Looks good. If it gets any more complex, I'd suggest using `argparse`, though at that point we should be refactoring the `scrape.yml` file into one per scraper.

PR #116, opened 2020-11-19
Issue 115 - Fixing issued dates issue

RedWalters:
Replaced hard xpath for dates, instead searching for correct path. Also has condition if no date is found to be able to continue with scrape without.

### DIFF
--- a/gssutils/scrapers/nrscotland.py
+++ b/gssutils/scrapers/nrscotland.py
@@ -43,8 +43,8 @@ def statistics_handler(scraper, tree):
                     try:
                         distribution.issued = parse(tree.xpath("//*[contains(text(),'Updated')]/text()")[0].lower().replace('last updated:', '')).date()
                     except:
-                        distribution.issued = '1900-01-01'
-                        logging.warning("No Last Issue Date Found, placement date provided (1900-01-01). Please update manually")
+                    	logging.warning("No Last Issue Date Found. Please update manually")
+                        pass                        
                 distribution.mediaType = {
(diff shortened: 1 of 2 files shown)


gssutils/scrapers/nrscotland.py
```diff
@@ -32,13 +36,15 @@ def statistics_handler(scraper, tree):
...
-                except:
-                    distribution.issued = parse(tree.xpath('//*[@id="node-stats-home-page-3066"]/div[2]/div/div/div[1]/table/tbody/tr/td[1]/div[2]/text()')[0].strip().lower().replace('last updated:', '')).date()
                 if 'Last update' in tree.xpath('//*[@id="block-system-main"]/div/div/div/div[2]/div/div/p[1]/text()'):
                     distribution.issued = parse(
                         tree.xpath('//*[@id="block-system-main"]/div/div/div/div[2]/div/div/p[1]/text()')[0]).date()
+                else:
+                    try:
+                        distribution.issued = parse(tree.xpath("//*[contains(text(),'Updated')]/text()")[0].lower().replace('last updated:', '')).date()
+                    except:
+                        distribution.issued = '1900-01-01'
```
mikeAdamss, line 46: with this (and the other one) I'd just leave them out. Throw a warning so people know they need to sort out the issued date but don't set it for them (cos they'll forget/ignore the warning). Just 

```
except:
    pass
```

the DE in question can sort it out int the transform.

mikeAdamss (approved):
makes sense :+1:

PR #118, opened 2020-11-20
Cubes/iteration 1

mikeAdamss:
cubes, first iteration. Encapsulate the csv, csvw and trig generation in a single convenient class. 

This is a hard one to read as there's a lot of ripping out of old ideas from before we're really nailed down what we wanted. To review I'd just look at cubes.py as a new (much simplified) class.

I've also dumbed down the tests significantly, they need doing proeprly, but the dependence on env vars and cwd() for url namespacing puts them out of scope for the timeframe in question.


ajtucker (approved):
This looks grand for iteration 1. We can fiddle under the hood as we get feedback and hopefully keep the API the same.

PR #125, opened 2020-12-10
move trace info to /out folder

mikeAdamss:
shift the trace json to its own file in /out rather than writing it back to info.json


ajtucker (approved):
Looks grand.

PR #126, opened 2020-12-14
Issue 94 measure type column

ajtucker:
Implement the "measure type" column approach by declaring a map for the measure type (and unit), e.g.:

```
      "Measure Type": {
        "dimension": "http://purl.org/linked-data/cube#measureType",
        "value": "http://gss-data.org.uk/def/measure/{measure_type}",
        "types": ["net-mass", "gbp-total"]
      },
      "Unit": {
        "attribute": "http://purl.org/linked-data/sdmx/2009/attribute#unitMeasure",
        "value": "http://gss-data.org.uk/def/concept/measurement-units/{unit}"
      },
      "Value": {
        "datatype": "integer"
      }
```

Note that

1) the `"types"` are provided so that we don't have to iterate through the whole CSV file to find the unique values for the measure type;
2) the `"Value"` column is identified as not having a `"measure"` or `"unit"` declaration;
3) the `"Value"` column can have only one datatype using this approach.

### DIFF
--- a/features/csvw.feature
+++ b/features/csvw.feature
@@ -121,3 +121,35 @@ Feature: Create CSVW metadata
     When I create a CSVW file from the mapping and CSV
     Then the metadata is valid JSON-LD
     And the input format of the csv is recorded as csv
+
+  Scenario: CSVMetadata from mapping with measure type and unit columns
+    Given a CSV file 'observations.csv'
+      | Period          | Flow    | HMRC Reporter Region | HMRC Partner Geography | SITC 4 | Value | Measure Type | Unit          |
+      | quarter/2018-Q1 | exports | EA                   | A                      | 01     | 2430  | net-mass     | kg-thousands  |
+      | quarter/2018-Q1 | exports | EA                   | A                      | 02     | 2     | net-mass     | kg-thousands  |
+      | quarter/2018-Q4 | imports | ZB                   | TR                     | 88     | 10    | gbp-total    | gbp-thousands |
+      | quarter/2018-Q4 | imports | ZB                   | TR                     | 89     | 352   | gbp-total    | gbp-thousands |
+    And a column map
+    """
+    "columns": {
+      "Period": {
+      },
+      "Flow": {
+      },
+      "SITC 4": {
+      },
+      "Value": {
+        "measureRef": "Measure Type",
+        "unitRef": "Unit",
+        "datatype": "integer"
+      }
+    }
+    """
+    And a dataset URI 'http://gss-data.org.uk/data/gss_data/trade/hmrc_rts'
+    When I create a CSVW file from the mapping and CSV
+    Then the metadata is valid JSON-LD
+    And gsscogs/csv2rdf generates RDF
+    And the RDF should pass the Data Cube integrity constraints
+    And the RDF should contain
+    """
+    """
--- a/features/csvw.feature
+++ b/features/csvw.feature
@@ -144,9 +144,16 @@ Feature: Create CSVW metadata
         "dimension": "http://gss-data.org.uk/def/dimension/sitc-4",
         "value": "http://gss-data.org.uk/def/concept/sitc-4/{sitc_4}"
       },
+      "Measure Type": {
+        "dimension": "http://purl.org/linked-data/cube#measureType",
+        "value": "http://gss-data.org.uk/def/measure/{measure_type}",
+        "types": ["net-mass", "gbp-total"]
+      },
+      "Unit": {
+        "attribute": "http://purl.org/linked-data/sdmx/2009/attribute#unitMeasure",
+        "value": "http://gss-data.org.uk/def/concept/measurement-units/{unit}"
+      },
       "Value": {
-        "measureRef": "Measure Type",
-        "unitRef": "Unit",
         "datatype": "integer"
       }
     }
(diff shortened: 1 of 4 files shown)


gssutils/csvw/mapping.py
```diff
@@ -229,6 +247,7 @@ def _as_csvw_object(self):
                                 rdfs_range=Resource(at_id=URI("http://purl.org/linked-data/cube#MeasureProperty"))
                             )
                         ),
+                        # ToDo: figure out the measures used, either from CSV column values, or declared in mapping.
```
robons, line 250: TO be removed?

features/csvw.feature
```diff
@@ -121,3 +121,167 @@ Feature: Create CSVW metadata
...
     Then the metadata is valid JSON-LD
     And the input format of the csv is recorded as csv
+
+  Scenario: CSVMetadata from mapping with measure type and unit columns
+    Given a CSV file 'observations.csv'
+      | Period          | Flow    | HMRC Reporter Region | HMRC Partner Geography | SITC 4 | Value | Measure Type | Unit          |
+      | quarter/2018-Q1 | exports | EA                   | A                      | 01     | 2430  | net-mass     | kg-thousands  |
+      | quarter/2018-Q1 | exports | EA                   | A                      | 02     | 2     | net-mass     | kg-thousands  |
+      | quarter/2018-Q4 | imports | ZB                   | TR                     | 88     | 10    | gbp-total    | gbp-thousands |
+      | quarter/2018-Q4 | imports | ZB                   | TR                     | 89     | 352   | gbp-total    | gbp-thousands |
```
robons, line 131: Shouldn't the measure types be `net mass` and `monetary value` (or similar)?

gssutils/csvw/mapping.py
```diff
@@ -122,6 +124,23 @@ def _validate(self):
             logging.error(f"Unknown prefixes used: {used_prefixes.difference(prefix_map.keys())}")
 
     def _as_csvw_object(self):
+        # Look to see whether the measure type has its own column
```
robons, line 127: The programmer in me says it's time to start breaking `_as_csvw_object()` down in to functions.

gssutils/csvw/mapping.py
```diff
@@ -256,6 +275,17 @@ def _as_csvw_object(self):
                         propertyUrl=URI("http://purl.org/linked-data/cube#measureType"),
                         valueUrl=URI(obj["measure"])
                     )
+                elif "datatype" in obj and not ("measure" in obj or "unit" in obj):
```
robons, line 278: Since the code which deals with `measure` and `unit` is above here, I would argue that if either of them are in this object containing `datatype` then we'd probably want an exception here to tell us what's wrong.

I'm all for failing hard when we have an inconsistency.

gssutils/csvw/mapping.py
```diff
@@ -256,6 +275,17 @@ def _as_csvw_object(self):
...
+                    # Where a measure type column exists
+                    assert self._measureTemplate is not None, "Must have a measure type column."
+                    self._columns[name] = self._columns[name]._replace(
+                        propertyUrl=self._measureTemplate.uri,
+                        datatype=obj["datatype"]
+                    )
+            elif self._mapping is not None and name in self._mapping and isinstance(self._mapping[name], bool):
+                self._columns[name] = self._columns[name]._replace(
+                    suppressOutput=not self._mapping[name]
```
robons, line 287: So this allows us to suppress a column by putting `"Column name": false` in the info.json transform section?

gssutils/csvw/mapping.py
ajtucker: We still need to remember to do this. I'll add a separate ticket.

features/csvw.feature
ajtucker: Yes. I was just basing this off some existing Tidy data and avoiding changing things as it's not part of this issue. However, I agree that it's distracting especially if this example ends up being canonical: we don't want to propagate the errors.

gssutils/csvw/mapping.py
ajtucker: Definitely. It's difficult to know when to stop refactoring and when to just get on with passing the tests, but agreed that this needs a good refactor.

gssutils/csvw/mapping.py
ajtucker: Aye. This is one of a few different ways I could think of doing this and still feels a bit dodgy. Failing fast is good.

gssutils/csvw/mapping.py
ajtucker: Yes. This is something I toyed with doing for the `Measure Type` and `Unit` mapping, giving the responsibility for them to the `Value` mapping and just suppressing them. I've found that there are other CSV files where extra columns are provided that could be suppressed, but which for instance might be used in a template.

gssutils/csvw/mapping.py
mikeAdamss: Agree with Rob (I like code that screams at me when things go wrong) but it feels like part of a/the future refactor. 

For me, every column "thing" should be mapped to a "type" up front, then errors are thrown for any and all oddities before we get on to actually doing anything (or similar, I'm sure everyone has their own thoughts). Validation of input I guess.

mikeAdamss (approved):
Some really interesting stuff here.

I think we all agree we need step back and refactor given how the scope of this class has crept since inception, but given we have a whole lot of blocked pipelines (and since it'd be "under the hood") I reckon we're safe to defer that for now - though a card would probably be a good idea.

Have approved, nice one Alex 👍

Issue #128, opened 2020-12-15
Catch errors for scraper Markdown preview


mikeAdamss:
the scraper markdown preview will currently throw an error if an expected attribute is missing.

we need to update so it catches the error and just lets the DE know if eg mediaType is missing, so (where viable) they can adapt the script without needing to update and redeploy gssutils to compensate.

this: https://github.com/GSS-Cogs/gss-utils/blob/7ec4fe7cd00c3dcea127ef9b28eefd85f1e146d9/gssutils/scrape.py#L101

ajtucker:
I know you said you'd added this issue, @mikeAdamss , but somehow I missed it when I created #129  :(

Meanwhile, I've added a PR.

PR #130, opened 2020-12-15
Guard against missing mediaType attribute

ajtucker:


ajtucker:
Sort of duplicates #128

mikeAdamss (commented):
Code is fab but I think we either need a catch for title and downloadURL or its not really a duplicate of 128. We've had pipelines break on _repr for lack of both before.

mikeAdamss (approved):
looks good

PR #133, opened 2020-12-29
Code List Manager

robons:
I've created a script which can be used to help manage the metadata associated with codelists. Enables users to create codelists by pointing the script at a raw CSV as well as updating existing `.csv-metadata.json` schema files. I've developed this to ensure that we can automatically add the necessary PMDCAT and DCAT metadata to them.

Installation instructions:
```
  pip install --user git+https://github.com/GSS-Cogs/gss-utils@ae0929dd8a6b25061754daee5a88d60a50ce86be#egg=gssutils
```

Ensure that your PATH variable contains `~/.local/bin` (might be a good idea to add this to your `~/.bash_profile` script).

Then you can go to any directory and run the `codelist-manager` script.

```
localhost:codelists HA9WF3CZ$ codelist-manager --help
usage: codelist-manager [-h] [-a AUTO] [-c CSV] [-s SCHEMA] [-u UPGRADE_ALL]

optional arguments:
  -h, --help            show this help message and exit
  -a AUTO, --auto AUTO  Automatically process upgrades without human input.
  -c CSV, --csv CSV     A CSV file. The script will create a corresponding
                        `.csv-metadata.json` file.
  -s SCHEMA, --schema SCHEMA
                        An individual schema file to upgrade.
  -u UPGRADE_ALL, --upgrade-all UPGRADE_ALL
                        Finds all `.csv-metadata.json` files within the
                        current directory (it recursively searches through
                        sub-directories) and applies any upgrades required.
```

Upgrade an existing `.csv-metadata.json` file:
```
localhost:codelists HA9WF3CZ$ codelist-manager -s species-group.csv-metadata.json 
Processing code list species-group.csv
Please provide the license URI: 
...
```

Generating a metadata file for an existing CSV:
```
localhost:codelists HA9WF3CZ$ codelist-manager -c admin-port.csv 
Is the code list defined at the Global level, the Family level or the Dataset level? (G/f/d): d
Please enter the family name (e.g. trade): chicken
Processing code list admin-port.csv
Please provide the license URI:
...
```

Created as part of https://github.com/GSS-Cogs/pmd-jenkins-library/issues/36


ajtucker (approved):
I've only had chance for a first pass over this but as it's critical to have codelists in PMD and the code doesn't impact existing code, think we should merge and start using ASAP.

Points for future:
* We're using `pathlib` (since Python 3.4) to manage file and path names, so should refactor to use this. I see e.g. path.split() being used which is always supposed to return a pair, so not sure why you have to guard against it returning a singleton.
* We've set up a way to do the Python OO -> RDF transformations in the metadata/*.py classes, so e.g. have pmdcat.CatalogRecord etc. Further, in e.g. https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/csvw/t2q.py#L210, we can convert this to JSON-LD.
* I'm uncomfortable putting in PMDCat or even catalogue metadata into the CSV-W metadata files. The responsibility for managing this catalogue level metadata should lie somewhere else.
* It would be good to figure out a way to avoid hard-coding URIs and using the base URI of the published CSV/CSV-W.

robons:
> I've only had chance for a first pass over this but as it's critical to have codelists in PMD and the code doesn't impact existing code, think we should merge and start using ASAP.
> 
> Points for future:
> 
>     * We're using `pathlib` (since Python 3.4) to manage file and path names, so should refactor to use this. I see e.g. path.split() being used which is always supposed to return a pair, so not sure why you have to guard against it returning a singleton.

The function recursively searches through parent directories for an `info.json` file. This means that if it doesn't find the file then eventually it reaches the root path - this is what's being guarded against so we don't throw an exception.

>     * We've set up a way to do the Python OO -> RDF transformations in the metadata/*.py classes, so e.g. have pmdcat.CatalogRecord etc. Further, in e.g. https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/csvw/t2q.py#L210, we can convert this to JSON-LD.
> 
>     * I'm uncomfortable putting in PMDCat or even catalogue metadata into the CSV-W metadata files. The responsibility for managing this catalogue level metadata should lie somewhere else.
> 
>     * It would be good to figure out a way to avoid hard-coding URIs and using the base URI of the published CSV/CSV-W.

I think the last two points are something we can discuss in the `info.json` workshop later today. I think it might be a good idea to provide the metadata required for this in the `info.json` file (or similar) and generate the `.csv-metadata.json` files in Jenkins. It'd made it much easier to edit and would mean we didn't have to have these URIs everywhere.

PR #134, opened 2021-01-04
Allow setting Codelist Location URI

robons:
SDMX/qb Dimensions must have codelists associated with the dimension property. We currently have no way of specifying where a dimension's codelist exists if it isn't a codelist local to the dataset.

This change is required to ensure we cabn pass the [pmd4/dsd.Dimensions Must Have Concepts In Code List](https://github.com/GSS-Cogs/gdp-sparql-tests/blob/c42c550197b1bc585fe667c4ec4c2e55c1f1784e/tests/pmd/pmd4/dsd/SELECT_DimensionsMustHaveConceptsInCodeList.sparql) SPARQL test.

https://github.com/GSS-Cogs/pmd-jenkins-library/issues/37

### DIFF
--- a/features/csvw.feature
+++ b/features/csvw.feature
@@ -108,7 +108,7 @@ Feature: Create CSVW metadata
 
       <#dimension/area-code> a qb:DimensionProperty ;
         rdfs:label "Area"@en ;
-        qb:codeList <#scheme/area-code> ;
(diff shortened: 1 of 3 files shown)


rossbowen (approved):

ajtucker:
This should be fine, though I had intended to use a registry (SPARQL queries over PMD instance) to look up bits of the DSD that should really be external. We should discuss where we can draw the line and keep things DRY.

PR #135, opened 2021-01-04
Allow undefined codelists for dimensions

robons:
This code change is to ensure that we can avoid setting codelists against dimensions where we there are continous ranges which simply cannot be represented by codelists.

See https://github.com/GSS-Cogs/pmd-jenkins-library/issues/37#issuecomment-754085987 for brief discussion.

### DIFF
--- a/features/csvw.feature
+++ b/features/csvw.feature
@@ -135,7 +135,8 @@ Feature: Create CSVW metadata
       "Period": {
         "label": "Quarter",
         "parent": "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod",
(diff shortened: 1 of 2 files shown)


ajtucker (approved):
Code looks fine, happy to approve.

I'm not sure about the principle however -- just need some discussion. E.g. IC-5 "Every dimension with range skos:Concept must have a qb:codeList" => refPeriod range shouldn't be a `skos:Concept`, which is fine, though I was wondering whether we still might be able to make statements about the range of (sub)properties of these types of properties (e.g. with continuous/infinite ranges) and use that for consistency checking.

PR #136, opened 2021-01-05
Fix Codelist URI in Multi-dataframe Datasets

robons:
Where a Dataset has multiple dataframes, gss-utils was previously incorrectly specifying the URI of codelists as relative to each dataframe, e.g. `http://gss-data.org.uk/data/gss_data/trade/hmrc_rts/first_dataframe#scheme/flow`. This PR corrects this so that the URIs are relative to the dataset e.g. `http://gss-data.org.uk/data/gss_data/trade/hmrc_rts#scheme/flow`.

Completed as part of https://github.com/GSS-Cogs/pmd-jenkins-library/issues/37

### DIFF
--- a/gssutils/csvw/mapping.py
+++ b/gssutils/csvw/mapping.py
@@ -6,6 +6,7 @@ from io import TextIOBase
 from pathlib import Path
 from typing import List, Optional, Dict, TextIO, Any, Set, Union
 from urllib.parse import urljoin
+import re
 
 from uritemplate import variables, URITemplate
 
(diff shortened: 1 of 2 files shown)


gssutils/csvw/mapping.py
```diff
@@ -51,16 +52,37 @@ def namify(column_header: str):
...
+            
+        Codelists are defined at the `dataset_root_path` level, so we need to be able to create URIs relative to
+            http://gss-data.org.uk/data/gss_data/<family_path>/<dataset_root_path>
+            
+        This function helps extract the URI up to the `dataset_root_path` fragment by removing any `dataset_path` 
+        fragment from `self._dataset_uri`.
+        """
+        if self._dataset_uri is None:
+            return None
+        matches: re.Match = re.match("^(.+)/gss_data/([^/]+)/([^/]+).*$", self._dataset_uri, re.RegexFlag.IGNORECASE)
```
robons, line 68: I don't like having to hardcode `gss_data` in here as this will break if we ever change the URI where datasets are located (or the Jenkins job name).

This is the simplest approach I could come up with without having to go back and edit every single main.py file to add a new param on to the CSVWMapping object.

ajtucker: I guess it's building up the dreaded technical debt, so we should schedule a chat about ways to address this in future and tag this as TODO.

How does this interact with @mikeAdamss 's `Cubes()` class?

robons: Well, this technical debt helps alleviate other technical debt so I think it's a step in the right direction.

So long as @mikeAdamss `Cubes` class is following the standard convention for building up URIs for each dataframe then there should be no problem.

rossbowen: Was there a reason for `gss_data` featuring so frequently in our uris historically? Always felt a little out of place to me!

ajtucker: The `gss_data` path was an accident. It's part of the Jenkins path to the job and I tried to strip it off at one point, but failed.

Happy for this to go through as long as we pin it with a TODO somewhere 👍

gssutils/csvw/mapping.py
```diff
@@ -94,9 +111,39 @@ def set_additional_foreign_key(self, foreign_key: ForeignKey):
...
+
+        If the dataset has more than one dataframe associated and so has a {uri} of the form
+            `http://gss-data.org.uk/data/gss_data/<family-name>/<dataset-name>/<dataframe-name>`
+        then the {dataset_root_uri} must represent the URI fragment common to all dataframes, e.g.
+            `http://gss-data.org.uk/data/gss_data/<family-name>/<dataset-name>`
+        """
         self._dataset_uri = uri
 
+        if dataset_root_uri is None:
+            print("WARNING: dataset_root_uri is unset. " +
```
robons, line 130: @ajtucker @rossbowen I've added an optional `dataset_root_uri` param which can be set when the dataset URI is set. It currently gives a warning when called, so hopefully developers will have a look and figure out what they need to set going forwards.

gssutils/csvw/mapping.py
robons: Could you approve it please? I can't merge until someone else does.

rossbowen (approved):

PR #139, opened 2021-01-06
Fixing bug where `value` URI is ignored

robons:
Unfortunately, due to the convoluted nature of code in the `CSVWMapping` class, interesting values in the info.json get ignored in unexpected circumstances. 

I'm fixing one of them so we listen to the `value` specified for a given column, instead of overriding it and pretending it's available locally.

### DIFF
--- a/gssutils/csvw/mapping.py
+++ b/gssutils/csvw/mapping.py
@@ -179,7 +179,7 @@ class CSVWMapping:
             # Codelist should exist. Convention dictates it should be a local codelist.
             return get_conventional_local_codelist_scheme_uri(column_name)
 
-        def get_convential_local_codelist_concept_uri_template(column_name: str) -> URI:
+        def get_conventional_local_codelist_concept_uri_template(column_name: str) -> URI:
             return self.join_dataset_uri(f"#concept/{pathify(column_name)}/{{{self._columns[column_name].name}}}",
                                          use_true_dataset_root=True)
 
@@ -187,7 +187,7 @@ class CSVWMapping:
             if "value" in column_def:
                 return URI(column_def["value"])
 
-            return get_convential_local_codelist_concept_uri_template(column_name)
+            return get_conventional_local_codelist_concept_uri_template(column_name)
 
         # Look to see whether the measure type has its own column
         for map_name, map_obj in self._mapping.items():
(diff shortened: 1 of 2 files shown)


gssutils/csvw/mapping.py
```diff
@@ -252,7 +262,7 @@ def get_maybe_codelist_for_col(column_config: object, column_name: str) -> Optio
                     self._keys.append(self._columns[name].name)
                     self._columns[name] = self._columns[name]._replace(
                         propertyUrl=self.join_dataset_uri(f"#dimension/{pathify(name)}"),
-                        valueUrl=self.join_dataset_uri(f"#concept/{pathify(name)}/{{{self._columns[name].name}}}")
```
robons, line 255: It used to ignore any `value` you specified here. I want it to use it please!

gssutils/csvw/mapping.py
```diff
@@ -177,7 +177,17 @@ def get_maybe_codelist_for_col(column_config: object, column_name: str) -> Optio
                 return Resource(at_id=codelist)
 
             # Codelist should exist. Convention dictates it should be a local codelist.
-            return get_conventional_local_codelist_uri(column_name)
+            return get_conventional_local_codelist_scheme_uri(column_name)
+
+        def get_convential_local_codelist_concept_uri_template(column_name: str) -> URI:
```
rossbowen, line 182: Typo (I think?) `convential` as opposed to `conventional`.

gssutils/csvw/mapping.py
```diff
@@ -344,7 +354,7 @@ def get_maybe_codelist_for_col(column_config: object, column_name: str) -> Optio
                 self._keys.append(self._columns[name].name)
                 self._columns[name] = self._columns[name]._replace(
                     propertyUrl=self.join_dataset_uri(f"#dimension/{pathify(name)}"),
-                    valueUrl=self.join_dataset_uri(f"#concept/{pathify(name)}/{{{self._columns[name].name}}}")
+                    valueUrl=get_convential_local_codelist_concept_uri_template(name)
```
rossbowen, line 357: `convential`

rossbowen (changes requested):

ajtucker (approved):
Looks grand. Can you exhibit the bug by updating the feature file with some expectations?

robons:
> Looks grand. Can you exhibit the bug by updating the feature file with some expectations?

I was hoping to avoid doing this, but I've committed a test now.

PR #142, opened 2021-01-12
remove cubes reliance on jenkins JOB_NAME env var

mikeAdamss:
the logic is wrong in cubes in regards to building urls when run on jenkins. the logic was supposed to be:

* use the path on local
* use the jenkins JOB_NAME env var on jenkins

as it stands its using **both** when run on jenkins and malforming the urls.

since the JOB_NAMR and the path are functionally equivalent (and having different handling is why the problem went undetected so long) I've switched it to just always use the file path for building urls.


ajtucker (approved):
Looks good.

PR #143, opened 2021-01-12
Adding support for metadata files with multiple tables listed

robons:
Upgrading the codelist-manager so that it can add metadata to CSV-W json files which reference multiple tables. It inspects every table and only upgrades ones which have type `skos:ConceptScheme`.

I've also refactored the code to use python's `pathlib` instead of `os.path`.

### DIFF
--- a/features/codelist-manager.feature
+++ b/features/codelist-manager.feature
@@ -301,7 +301,7 @@ Scenario: Creating New Global Level Metadata File from CSV with atypical column 
     }
     """
 
-  Scenario: Handling multi-table schemas.
+Scenario: Handling multi-table schemas.
     Given We have a CSV file named "my-favourite-code-list.csv" with headers
       """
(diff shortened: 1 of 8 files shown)


ajtucker (approved):
This looks fine and dandy.

I'd like to consider, separately, re-using the named-tuple approach to encapsulating the CSV-W table model, as per https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/csvw/table.py, though we'd need to factor out the `as_plain_obj()` method from https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/csvw/mapping.py#L414, perhaps as a mixin.

Actually, one thing I miss from Scala is pattern matching with the `apply()` and `unapply()` methods. http://stupidpythonideas.blogspot.com/2015/12/pattern-matching-again.html has some interesting ideas and now we have some more type hints it might make things nicer.

PR #147, opened 2021-01-18
Python Pandas Upgrade

canwaf:
Merged in master to finish the work; however Python bumped to 3.9, pandas version unleashed.


ajtucker (approved):
It builds and all bar one of the tests pass and that test is now out of date, so I think we should merge to master, tag a new release and have a go locally with some pipelines.

PR #148, opened 2021-01-20
Api scraper mike

mikeAdamss:
progress so far. Some tweaks to line stuff up and about half of the tests are in (at least partially).

could be worth merging in so we can write against the ones that work, unsure.


gssutils/metadata/dcat.py
```diff
@@ -300,11 +312,22 @@ def get_odata_api_periods(distro: Distribution) -> list:
...
+    return formatted_periods
+
+
+def format_as_url_period(period: str) -> str:
+    """
+    Given a period of time formatted as per odata conventions, convert to a full
+    period url as used by pmd4
+    """
+
+    # TODO - this function probably needs to be "convert_to_datetime"
```
canwaf, line 328: The pattern here would be convert a string to datetime then that datetime to a string. 

```
pd.to_datetime(df['Period'].str[:4], format='%Y').dt.strftime('/id/year/%Y')
pd.to_datetime(df['Period'], format='%Y%b').dt.strftime('/id/month/%Y-%m')
```

But is this necessary? Is our goal to have OData API JupyterNotebooks be

1. Initialise scraper, cubes, read in info.json
2. Set `cube = scraper.dataset(latest=True).as_pandas()`
3. Add cube to cubes
4. `cubes.output_all()`

I envisaged a little more motion at 2 to allow for gluing things together, specifically to allow the designer to translate the url_periods automatically, and possibly put their finger on the scales to decide which periods get called.

mikeAdamss: I'll add in passing kwargs to our function, that's seems like a grand idea anyway and requesting periods from there should be easy enough.

RE the datetime angle, I was thinking more in terms of having a common currency, is there potential for a mix of eg weeks, months and years in the returned data? getting everything into datetime was aimed at that possibility, you can find the "latest" between two datetime objects easier than between eg one string of week and one string of month.

I'm sketchy on whether that holds btw so no issues if it doesn't. 

When you say finger on the scale what's in your head?

canwaf: After review not required.

gssutils/metadata/dcat.py
```diff
@@ -300,11 +312,22 @@ def get_odata_api_periods(distro: Distribution) -> list:
...
     formatted_periods = []
     for period in periods:
-        
-        # format as per pmd
-        year = str(period)[:4]
-        month = str(period)[-2:]
-        period = f'/month/{year}-{month}'
+
+        period = format_as_url_period(period)
         formatted_periods.append(period)
```
canwaf, line 318: Not required.

canwaf (changes requested):
Removing of items not in MVP based on comments, and associated tests can go.

Summary of what the main.py will look like after we're done.

```initialise everything +
- cubes +
- scraper +

compare endpoints +
- get periods on odata as operiods *
- get periods on sparql as speriods * 
- translate operiods to speriods +
- pass speriods and operiods to compare !
  - save result as next_period +

get data +
- dataset.as_pandas(period=next_period) *
- sup_dfs.fetch() *

complete dataset +
- merge() !

finish +
- output everything after adding to cubes +

* is MVP
! is stuff we will do, but not MVP
+ is manual stuff which will exist in dataset's main.py
```

canwaf (approved):

PR #162, opened 2021-01-26
Api scraper mike

mikeAdamss:
there's a lot of code changes here because of the refactoring but it's basically.

* move the download functions out of metadata.dcat.py and into transform.download.py.
* finish (most of) the steps to get all tests running
* tweaks to get tings working, plus expose the time getting function for import from gssutils
* record the http responses so we can run tests without hammering api's.

### Needs doing
* the longer living cache, added a few bits but needs to be done properly and test needs to be real
* correct tests - they all run ..but... most have missing fixtures or (possibly, not sure) need a logic tweak.
* the imports are a wreck from all the refactoring, need to clean it up a bit* 
....a million things I'm not thinking of

from here we should be able to run the individual tests, get them working/fixed and begin pushing things over the finish line.


canwaf (approved):
Build fails for other reasons, will check locally, but merging into another branch won't be an issue.

PR #164, opened 2021-01-27
Add OData API EndPoint functionality to as_pandas()

canwaf:
This new functionality is heavily dependent on an extended info.json (though the ultimate location where this ends up is entirely up to Data Managers).

1. Initialise an API-based dataset as usual
```
infoFileName = 'features/fixtures/seed-for-api-scraper.json'

info    = json.load(open(infoFileName))
scraper = Scraper(seed=infoFileName)
cubes   = Cubes(infoFileName)
```
2. Set the distribution (which makes accessing features easier) `distro = scraper.distribution(latest=True)`
3. Get the api_chunks `api_chunks = get_odata_api_periods(distro)`
4. Get the pmd_chunks `pmd_chunks = get_pmd_periods(distro)`
5. Determine how which dataset to download next using set math and taking the minimum like `min(api_chunks-pmd_chunks)`. It is *highly likely* the user will have to translate between api chunks and pmd_chunks, and this item isn't bootstrapped, so knowledge of how the data will be represented in PMD is required before.
6. The full dataset with all its columns (many of which may not be useful) chunked to the chunk parameter you determined comes from `df = distro.as_pandas(chunks_wanted=list or string or int)`
7. That's it

The format for the `info.json` additions is so far as follows:
```
  "odataConversion": {
      "publishedLocation": "https://staging.gss-data.org.uk/sparql",
      "datasetIdentifier": "http://gss-data.org.uk/data/gss_data/trade/hmrc-ots#dataset",
      "chunkColumn": "MonthId",
      "chunkDimension": "dim:refPeriod",
      "supplementalEndpoints": {
          "FlowType": {
              "endpoint": "https://api.uktradeinfo.com/FlowType",
              "primaryKey": "FlowTypeId",
              "foreignKey": "FlowTypeId"
          }
  }
```

* `publishedLocation` is the SPARQL endpoint where the data ultimately ends up
* `datasetIdentifier` is the `qb:dataset` fully qualified URL
* `chunkColumn` is the API's column you're going to chunk on
* `chunkDimension` is the PMD dimension you're going to chunk on
* `chunkColumn` and `chunkDimension` need to be equivalent, but also translated in the notebook by the Data Engineer
* `supplementalEndpoints` define the supplemental datasets which enhance the principle dataset from the principle datasets' `downloadURL`
* `endpoint` is which API EndPoint is going to be used to fetch it, and a primary/foreign key pair
* `primaryKey` is the primary key attribute of the supplemental dataset which is used to merge the data to the principle dataset
* `foreignKey` is the attribute of the principle dataset which links to the primary key in the supplemental dataset
* The name of the `supplementalEndPoints` do not matter, provided they are unique to the `info.json`

This is because it is easier to build the MVP against a malfunctioning OData API (i.e. the uktradeinfo API doesn't have a complete object model) then have the routine scrape the object model from the API. That is for another time.


ajtucker (approved):
I've encapsulated the downloads functionality into a mixin `Downloadable` class. It's not quite right, but good enough for now. It does highlight that the BDD tests are looking inside the implementation of the class.

The download steps and features are similar to the distribution steps and features, so we should look at refactoring and amalgamating.

I'm uncomfortable with the responsibility of interrogating the downstream PMD to see what it contains. We need to have a think about the process for chunking up datasets, but that's outside the immediate concerns of this PR.

Let's approve this and make some time to work out what approaches we can take to keeping datasets up to date without having to download the whole lot every time.

PR #167, opened 2021-02-01
Fix/ons scraper modified date

mikeAdamss:
Update the ONS scraper to capture the modified date of each version of a dataset in a time seies (so we can filter to latest based on date modified where we want to).

so this kind of thing will be possible:

```
from gssutils.metadata.mimetype import Excel
latest_modified_version = max([x.modified for x in scraper.distributions])
distro = scraper.distribution(modified=latest_modified_version, mediaType=Excel)
```


canwaf (approved):

ajtucker (approved):
Looks good, though hard to tell without tests ;-)

PR #168, opened 2021-02-01
unencapsulating get_chunks methods

canwaf:
Just a small change to download.py both method and tests to remove the underscore prefix for get_odata_api_chunks and get_pmd_chunks, which now are accessible without passing the distro to it, as they are now distro functions which are called by `distribution(latest=True).get_odata_api_chunks`, etc.


ajtucker (approved):
👍

PR #176, opened 2021-02-05
Implement multiple families / themes

ajtucker:
Modify test to check output. Mostly deal with values being able to be singleton strings, or lists of strings.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ pipeline {
             agent {
                 dockerfile {
                     args '-u root:root'
+                    reuseNode true
                 }
             }
             steps {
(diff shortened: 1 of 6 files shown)


ajtucker:
Note that the RDF output from `Cubes()` is checked and that it's looking for `pmdcat:Dataset` rather than `dcat:Dataset`. We really should be just doing DCAT here.

ajtucker:
Hmm, [tests are failing](https://ci.floop.org.uk/job/GSS_data/job/gss-utils-branches/view/change-requests/job/PR-176/2/cucumber-html-reports/report-feature_3_3381053843.html) because the current working directory is different on my local machine to on Jenkins in the branch pipeline.

We should fix this dodgy coupling on the ID of the dataset and the name of the local directory, see GSS-Cogs/airtable-utils#11.

mikeAdamss (approved):
nice one 😄

PR #181, opened 2021-02-12
 Bug/empty ons distributions

mikeAdamss:
Where dataset versions field of the ONS /data endpoints, are only populated when 2 or more version of a dataset have been release, eg:

1.) 1 distribution has 0 versions
2.) 2 distributions has 2 versions
3.) 3 distributions has 3 versions

when we updated the handling to look for updatedDate (to fix release date bug) it doesn't account for scenario 1.

this pr adds a fallback, so where there are 0 version of a dataset, we use the principle page information instead.


ajtucker (approved):
Great stuff, thanks Mike!

PR #185, opened 2021-02-18
Bug/issue 184 - ONS Scraper missing latest/current version

mikeAdamss:
We're having issues still around ONS release dates.

On the api there's only a single release date (the date the very first version of the data was released) but every subsequent versions comes with an "updateDate" field, which rather than being the date _that_ version was last changed, is the date the dataset was **superceded** by the next one.

So this pr unpicks that to assign correct release dates to scraped distributions which will allow us to use our typical `scraper.distribution(latest=True)` patten with ons datasets.

### DIFF
--- a/features/scrape.feature
+++ b/features/scrape.feature
@@ -194,15 +194,18 @@ Feature: Scrape dataset info
     And select the distribution given by
       | key       | value                                                                    |
       | mediaType | application/vnd.ms-excel        |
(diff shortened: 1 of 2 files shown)


ajtucker (approved):
This looks grand.

It reminds me that we need to follow up the local date / timezone issue with ONS before daylight saving starts again.

PR #186, opened 2021-02-18
Move StatsWales to _get_odata_ functionality

canwaf:
Small changes:

simple odata dataframes now use the backoff decorated _get_odata_data, and now chunks are fully optional (to support simple odata data fetching).


gssutils/transform/download.py
```diff
@@ -162,6 +155,12 @@ def _get_odata_data(self, url: str, params: Optional[dict] = None) -> pd.DataFra
 
         contents = r.json()
         df = pd.DataFrame(contents['value'])
+
```
mikeAdamss, line 158: im guessing this is for the statswales odata one, it's probably worth adding a comment againt both these `while` statements, just so we "get it" six months from now.

mikeAdamss (commented):
code looks good but you should probably add a bdd scenario to represent this (hopefully you can reuse the existing steps, just with different arguments).

gssutils/transform/download.py
canwaf: Done.

canwaf:
> code looks good but you should probably add a bdd scenario to represent this (hopefully you can reuse the existing steps, just with different arguments).

I don't think this is necessary since the Stat Wales ones covers this behaviour off.

gssutils/transform/download.py
```diff
@@ -100,29 +100,22 @@ def _get_simple_csv_pandas(self, **kwargs) -> Union[Dict[str, pd.DataFrame], pd.
...
+            return self._get_principle_dataframe()
         raise FormatError(f'Unable to load {self._mediaType} into Pandas DataFrame.')
 
-    def _get_principle_dataframe(self, chunks_wanted: list = None):
+    def _get_principle_dataframe(self, chunks_wanted: Optional[list] = None):
         """
         Given a distribution object and a list of chunks of data we want
         return a dataframe
         """
         principle_url = self.uri
```
CharlesRendle, line 112: Not really the subject of this review.

If `principle_url` is defined here, why do we need `url` (line 130) when `chunks_wanted is None` ?
Definitely could be missing something obvious or just being needlessly pedantic!

CharlesRendle (commented):
Happy with the changes made for StatsWales `_get_odata_` functionality.

Happy that repeated statements across lines 160-163 and 166-169 are not posing too much of an issue which would take more lines to solve the current implementation as well as maintain readability of code.

gssutils/transform/download.py
canwaf: Got rid of the lot, tbh. No point assigning the variable.

mikeAdamss (approved):
looks good 👍

PR #189, opened 2021-02-23
catch for ons-scraper,  no initial release date

mikeAdamss:
set the ons scraper to ignore initial (v1) distributions that don't have a release date (rather than falling over as it'll currently do).

also tidied up some old exceptions, one was related to legacy approach to release dates we don't use anymore, the other was raising an exception and killing the scrape where a warning would have done.


ajtucker (approved):
Nice one. Looks good!

PR #200, opened 2021-03-12
Get get_pmd_chunks() to lie when AccretiveUploads is false

### LINKED ISSUE #199: Get get_pmd_chunks() to lie when AccretiveUploads is false
A dataset may need to be restarted from scratch and replace the published data (i.e. when a code list is promoted to family-level etc.), right now the get_pmd_chunks() function will return the values from PMD and always tell the truth.

Jenkins now [has an environment variable](https://github.com/GSS-Cogs/pmd-jenkins-library/issues/53) which will temporarily change the AccretiveUpload key's value in `info.json` to false in order to replace the cube in its entirety on PMD. We need to respect this, but do so in a way which requires minimal changes to existing pipelines.

So the change is as follows, if AccretiveUpload is false, return a null set from PMD.

canwaf:
Fixes #199

### DIFF
--- a/gssutils/transform/download.py
+++ b/gssutils/transform/download.py
@@ -211,6 +211,10 @@ class Downloadable(Resource):
         Given the downloadURL from the scraper, return a list of chunks from pmd4
         """
 
+        if 'load' in self._seed and 'accretiveUpload' in self._seed['load'] and not self._seed['load']['accretiveUpload']: 
+            return list()
+
+
         # Assumption that no cases of multiple datasets from a single API endpoint, so...
         dataset_url = self._seed['odataConversion']['datasetIdentifier']
         endpoint_url = self._seed['odataConversion']['publishedLocation']


gssutils/transform/download.py
```diff
@@ -211,6 +211,10 @@ def get_pmd_chunks(self) -> list:
         Given the downloadURL from the scraper, return a list of chunks from pmd4
         """
 
+        if 'load' in self._seed and 'accretiveUpload' in self._seed['load'] and not self._seed['load']['accretiveUpload']: 
+            return list()
```
robons, line 215: I'd add a comment describing why this is here.

Something like `# If this is the case, the Jenkins pipeline is informing us that it is performing a drop & replace upload. We need to re-seed the data.`

robons (approved):

mikeAdamss:
late comer but this is a bit of an odd beast, does this need to be in gssutils at all?

if you absolutely have to have an env var of some sort telling you the a lsit should be empty, can we just do something like this in the main.py.

```
pmd_chunks = [] if os.environ.get("VAR_NAME", None) is not None else distro.get_pmd_chunks()
```

robons:
> late comer but this is a bit of an odd beast, does this need to be in gssutils at all?
> 
> if you absolutely have to have an env var of some sort telling you the a list should be empty, can we just do some variation of this in the main.py.
> 
> ```
> pmd_chunks = [] if os.environ.get("START_OVER") else distro.get_pmd_chunks()
> ```

I can certainly set an environment variable for you if you like.

mikeAdamss:
Ignore my mention of env var, getting my env vars and info.json properties mixed up it seems 😄

Guess my wider point is a method to return an empty list on demand seems a bit odd, wondering if we would be better off just checking if we want pmd chunks or not in the main.py.

canwaf:
The reason why you'd build this in gss-utils is that you'd have to build the check in main.py each time, and remember that it's part of the process... or abstract that process away from the pipeline and keep the pipeline focused on delivering the data required and have gss-utils manage how to what to build/deliver.

canwaf:
Hey @mikeAdamss, you happy with this push or should we consider another way to implement a solution to this problem?

mikeAdamss:
I can see the point. I guess my take is that in this scenario we built the `get_pmd_chunks` method specifically so the logic would be under the control of the data engineer writing the transform. Having that fuctionality of that method changed via an external variable seems at odds with that idea.

So like `get_pmd_chunks` returns pmd chunk as you'd expect ... except sometimes it doesn't. That feels like asking for trouble.

So it ain't wrong (as in, it'd work just grand) but I'm worried about a scenario either you know exactly how this works or it's going to be confusing af.

So I ain't saying never merge this (might end up as the best option) but I would personally mark it as a draft and explore alternatives. We can always fall back to this if need be.

canwaf:
The point of this is to be able to start the process again from scratch from Jenkins. But another solution would be to just delete the data on PMD for it to start again, or have the pipeline accept a variable which tells it which period to start with.

mikeAdamss:
Tbh, I think I'm failing to make my point or I'm misunderstanding what you're up to here.

Having a method thats behaviour silently changes due to external factors is typically to be avoided.

That doesn't mean don't aim for your goal here, it just means I'm not sure we want to implement it in quite this way.

Off the top of my head:
- you could use some sort of keyword, making it explicit (and obvious) that the method may return an empty list based on config.
- you could wrap the check for convenience, something vaguely like `distro.is_accretive(): pmd_chunks = <get the chunks> else [] `
- as an absolute minimum, you could log out a bloody massive warning telling us when we're changing the behaviour of that method behind the scenes.
- or you could just do it in the main,py

There's plenty of ways to skin this cat that are less likely to trip us up down the line, is the point Ive been trying to make.

canwaf:
Don't need this, workaround is in PMD.

robons:
> Don't need this, workaround is in PMD.

What's the workaround?

PR #201, opened 2021-03-12
Accretive Upload - Stop generating Trig and DSD

robons:
When we're doing an accretive upload, we need to ensure that the Data Structure Definition and the trig file containing catalogue metadata aren't generated so that we don't duplicate information which is already stored on PMD.

This isn't a problem with our standard uploads since they're done in a drop-and-replace fashion, but this has become an issue now that we're keeping the existing data and adding new qb:Observations to the dataset. 

Part of https://github.com/GSS-Cogs/pmd-jenkins-library/issues/53

### DIFF
--- a/gssutils/transform/cubes.py
+++ b/gssutils/transform/cubes.py
@@ -100,10 +100,10 @@ class Cube:
 
         # Use the info.json for the mapping by default, but let people
         # pass a new one in (for where we need to get clever)
-        if self.info_json_dict is None:
-            map_obj.set_mapping(info_json)
-        else:
-            map_obj.set_mapping(self.info_json_dict)
+        info_json = info_json if self.info_json_dict is None else self.info_json_dict
+
+        map_obj.set_accretive_upload(info_json)
+        map_obj.set_mapping(info_json)
 
         map_obj.set_csv(destination_folder / f'{pathified_title}.csv')
         map_obj.set_dataset_uri(urljoin(self.scraper._base_uri, f'data/{self.scraper._dataset_id}'))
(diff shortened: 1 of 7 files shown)


gssutils/transform/cubes.py
```diff
@@ -100,10 +100,10 @@ def _instantiate_map(self, destination_folder, pathified_title, info_json):
 
         # Use the info.json for the mapping by default, but let people
         # pass a new one in (for where we need to get clever)
-        if self.info_json_dict is None:
-            map_obj.set_mapping(info_json)
-        else:
-            map_obj.set_mapping(self.info_json_dict)
+        info_json = info_json if self.info_json_dict is None else self.info_json_dict
```
canwaf, line 103: Ooh. Fancy.

gssutils/transform/cubes.py
```diff
@@ -154,10 +154,16 @@ def output(self, destination_folder, is_multi_cube, is_many_to_one, info_json):
         # output the tidy data
         self.dataframe.to_csv(destination_folder / f'{pathify(self.title)}.csv', index=False)
 
-        # Output the trig
-        trig_to_use = self.scraper.generate_trig()
-        with open(destination_folder / f'{pathify(self.title)}.csv-metadata.trig', 'wb') as metadata:
-            metadata.write(trig_to_use)
+        is_accretive_upload = info_json is not None and "load" in info_json and "accretiveUpload" in info_json["load"] \
```
canwaf, line 157: Now this one is confusing, but fine.

canwaf (approved):

gssutils/transform/cubes.py
robons: What's confusing?

Issue #203, opened 2021-03-16
pmdcat:DataCube not expanding correctly


rossbowen:
Using `pmdcat` as a compact URI in JSON-LD does not expand without placing `pmdcat` in the `@context`. This results in triples such as 

```ttl
<http://gss-data.org.uk/data/trade/some-data#dataset> a <pmdcat:DataCube> .
```

appearing in our output.

https://github.com/GSS-Cogs/gss-utils/blob/734a6b1876a06bd5bc79e9f50a131a4235be58c9/gssutils/csvw/dsd.py#L67

Rather than placing in the `@context`, I prefer providing a fully qualified PMD DataCube URI, `http://publishmydata.com/pmdcat#DataCube`.

ajtucker:
See also https://github.com/GSS-Cogs/gss-utils/pull/171#issue-566755648

PR #204, opened 2021-03-16
Fix pmdcat:DataCube not expanding correctly

rossbowen:
See https://github.com/GSS-Cogs/gss-utils/issues/203.

### DIFF
--- a/features/csvw.feature
+++ b/features/csvw.feature
@@ -80,9 +80,8 @@ Feature: Create CSVW metadata
       @prefix sdmx-c: <http://purl.org/linked-data/sdmx/2009/code#> .
       @prefix gss-dim: <http://gss-data.org.uk/def/dimension/> .
       @prefix gss-meas: <http://gss-data.org.uk/def/measure/> .
-      @prefix pmdcat: <http://publishmydata.com/pmdcat#> .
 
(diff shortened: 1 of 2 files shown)


ajtucker:
It'd be good to have test coverage of this.

ajtucker:
I wonder whether the fix is to remove the declaration from `dsd.py`, as it's already declared properly in `pmdcat.py`. I think it's currently getting both `<pmdcat:DataCube>` and `pmdcat:DataCube` types, e.g. https://staging.gss-data.org.uk/resource?uri=http%3A%2F%2Fgss-data.org.uk%2Fdata%2Fgss_data%2Ftrade%2Fons-uk-trade-in-services-by-industry-country-and-service-type-exports%2Fpcn%23dataset

rossbowen:
I think you're right. It's appearing in the `.trig` which I think is coming from here:

https://github.com/GSS-Cogs/gss-utils/blob/734a6b1876a06bd5bc79e9f50a131a4235be58c9/gssutils/scrape.py#L384-L385

ajtucker (approved):
Thanks for tracking this down!

I added a minimal check for the right triple, though this won't guard against it being wrong.

Note that we're accreting technical debt :)

PR #206, opened 2021-03-16
get_pmd_chunks query is malformatted

### LINKED ISSUE #205: get_pmd_chunks query is malformatted
The query from PMD is formatted via f-string, and assumed that all chunk_dimensions would be a dim: prefix, which is not the case. So the following line

```
        query = f'''PREFIX qb: <http://purl.org/linked-data/cube#>
PREFIX dim: <http://purl.org/linked-data/sdmx/2009/dimension#>
SELECT DISTINCT ?chunk WHERE {{
    ?obs qb:dataSet <{dataset_url}>; {chunk_dimension} ?chunk .
}}'''
```

Needs to change to have the chunk dimension in `<>` brackets.

canwaf:
Fixes #205


canwaf:
This is a very minor change, which will fix the HMRC_RTS re-run problems @robons, plus changes to the HMRC_RTS `info.json`.

robons (approved):

PR #214, opened 2021-03-19
Split out scrape.yml

ajtucker:
Note that this hasn't cleaned out any duplicate requests that might have found their way in, but at least makes it easier to work on individual scrapers in isolation.


mikeAdamss (approved):
nice, looks good 👍

PR #216, opened 2021-03-22
get_odata_api_chunks() query is malformatted

### LINKED ISSUE #215: get_odata_api_chunks() query is malformatted
`apply` keyword requires `$` prior

canwaf:
Fixes #215

Workaround will be implmeneted in HMRC-regional-trade-statistics until this is fixed, the workaround is direct session access so it's not a big deal to remove and return to the old version.

Right now the bug returns 30,000 full records from the uktradedata api, where it should only return the month ids because the apply query was malformatted.


ajtucker (approved):
Looks to be on the money :$

PR #218, opened 2021-03-22
Issue 217/govuk guidance scraper

CharlesRendle:
Scraper for govuk guidance page plus test

Paired with Mike


ajtucker:
Does the content API, https://content-api.publishing.service.gov.uk/, work with these?

gssutils/scrapers/govuk.py
```diff
@@ -284,4 +287,48 @@ def eth_facts_service(scraper, tree):
             pass
 
 
+def content_api_guidance(scraper, tree):
```
mikeAdamss, line 290: this is a left over of the old pattern, but if you look up to line 40 the thing we're passing isn't `tree` anymore it's `metadata`. Which is you look at line 23 is already a dictionary of what we need 🎉

gssutils/scrapers/govuk.py
```diff
@@ -284,4 +287,48 @@ def eth_facts_service(scraper, tree):
...
+    
+    content_api_url = f'https://www.gov.uk/api/content/{scraper.uri.replace("https://www.gov.uk/", "")}'
+    r = scraper.session.get(content_api_url)
+    if r.status_code != 200:
+        raise Exception(f'Failed to get url "{content_api_url}", status_code "{r.status_code}".')
+    page_json = r.json()
```
mikeAdamss, line 296: this entire block (which I'll admit is my fault 😄 ) isn't needed anymore, Alex has covered it earlier on. What we were called `page_json` is exactly the same thing as the `metadata` dict being passed in (see above comment), you can remove all of this.

just switch any of the below references to `page_json` into `metadata` and you're golden.

gssutils/scrapers/govuk.py
```diff
@@ -284,4 +287,48 @@ def eth_facts_service(scraper, tree):
...
+
+    if 'title' in page_json:
```
mikeAdamss, line 298: this pattern ends up as a silent fail which we'd rather avoid really (as-in this would work if the title is missing - which is good, but we'd want it screaming at us telling us whats happening under the hood), so something like:
```
title = metadata.get("title", None)
if title is None:
    logging.warning(f'The title for  dataset {scraper.url} not set, title field missing from content api')
 else:
   dataset.title = title
```
etc for the other dataset level fields.

gssutils/scrapers/govuk.py
```diff
@@ -284,4 +287,48 @@ def eth_facts_service(scraper, tree):
...
+    if 'links' in page_json and 'organisations' in page_json['links']:
+        orgs = page_json['links']['organisations']
+        if len(orgs) == 0:
+            logging.warning("No publishing organisations listed.")
+        elif len(orgs) >= 1:
+            if len(orgs) > 1:
+                logging.warning('More than one organisation listed, taking the first.')
+            scraper.dataset.publisher = orgs[0]["web_url"]
+
+    for attachment in page_json['details']['attachments']:
```
mikeAdamss, line 321: I'd wrap this in a try catch, there's a better than average chance that somewhere on gov.uk they're adding some sort of weird attachment that'll break it, better to control getting a broken one (and again, shout about it), so something like:
```
try:
       <all the code in this loop>
except KeyError:
       # Note - "KeyError" happens if you ask a dict for something that ain't there, which covers any fails here
      logging.warning(f'Failed to extract attachment {json.dumps(attachment, indent=2)}'
```

you might need to import json, cant remember if we have.

mikeAdamss (commented):
this is genuinely fine, nice one (the only thing that'd actually break stuff is the unnecessary requests.get bit I added yesterday 😆)

a few comments but all around making the scraper more robust rather than any alarm bell stuff.

gssutils/scrapers/govuk.py
```diff
@@ -287,48 +288,67 @@ def eth_facts_service(scraper, tree):
...
         if len(orgs) == 0:
             logging.warning("No publishing organisations listed.")
         elif len(orgs) >= 1:
             if len(orgs) > 1:
                 logging.warning('More than one organisation listed, taking the first.')
             scraper.dataset.publisher = orgs[0]["web_url"]
 
-    for attachment in page_json['details']['attachments']:
-        distro = Distribution(scraper)
+    try:
```
mikeAdamss, line 328: we need to change the position of this try/except catch.

as it is, as soon as it throws as exception it'll drop down to the except (which is outside the loop), so it'll survive hitting a bad attachment ..but... it'll also stop scraping distributions, needs to be:

```
for attachment in metadata['details']['attachments']:
     try:
            # stuff
      except:
            logging. etc etc
```

that way if a bad attachment is attachment 2 of 3, you'll still scrape 3.

gssutils/scrapers/govuk.py
```diff
@@ -287,48 +288,67 @@ def eth_facts_service(scraper, tree):
...
+        for attachment in metadata['details']['attachments']:
+            distro = Distribution(scraper)
+
+            dist_title = attachment.get('title', None)
```
mikeAdamss, line 332: I'd let this (and especially the downloadURL) fail. If a distribution is missing one or both of title and downloadURL it's not much use to us anyway, better to let eg `dist.title = attachment.get('title')` fail with a KeyError and let the except catch handle it.

...to explain that.... we're allowing and logging the blank fields at the _dataset_ level because without a dataset to put distributions into you've got literally nothing, blank scraper, as there's no existant _thing_ to put the data into. 

Whereas not scraping a _distribution_ that's irrelevant to us (as we need a downloadURL and title) is probably completely fine, we do it all the time (I say _probably_ because you never know, but given we're loudly logging out the failed attachments with the catch you're safe to let the non-title non-downloadURL ones fail).

mikeAdamss (commented):
almost, almost there. Just added some comments around the wonderfully quirky world of exception handling 😄

gssutils/scrapers/govuk.py
CharlesRendle: Try/catch exception moved entirely within the loop.

gssutils/scrapers/govuk.py
CharlesRendle: Removed `None` components from `dist_title` and `dist_url`.

mikeAdamss (approved):
looks good to me, nice one 👍

Issue #160, opened 2021-01-25
codelistmanager gets wrong dataset URL for local codelists


ajtucker:
Our convention to date has been that the dataset URIs are formed from a base URI plus a path that includes the dataset family name, the publisher and the dataset title, e.g.:

```
http://gss-data.org.uk/data/gss_data/covid-19/wg-notifications-of-deaths-of-residents-related-to-covid-19-in-adult-care-homes-catalog-entry
```

The information required to build this URI is ultimately stored in the `datasets/info.json` / `pipelines` list and in the file path of the dataset, e.g. `WG-Notifications-of-deaths-of-residents-related-to-COVID-19-in-adult-care-homes`.

For completeness, we could repeat this identifier in the dataset's own `info.json` file, see https://github.com/GSS-Cogs/airtable-utils/issues/11

In any case, https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/codelistmanager/createnew.py#L226 needs updating to get the right dataset URI.

robons:
Upgrade deployed.

Upgrade with:
`pip install --user --upgrade 'git+https://github.com/GSS-Cogs/gss-utils#egg=gssutils'`

PR #227, opened 2021-03-29
Use the new info.json `id` field in preference to the `title` field.

robons:
Fix da bug.

https://github.com/GSS-Cogs/gss-utils/issues/160


ajtucker (approved):
Looks fine.

Issue #224, opened 2021-03-25
Key trade dimensions having subPropertyOf in place is critical to the current search implementation


rossbowen:

rossbowen:
## Intro

The Trade structural search navigates through a tree of dimensions using `rdfs:subPropertyOf` relationships.

For time period dimensions, or geographical dimensions, we commonly specify SDMX `refPeriod` or `refarea` dimensions as parents of our locally defined dimensions.

- In the interest of enabling the search in short term, we want to put sensible and logically correct relationships in place using `rdfs:subPropertyOf`.
- In doing this we want to continue to meet integrity constraints.

## Situation

A dataset  has a "product" dimension which primarily uses SITC codes but extends with a few bespoke codes.
A different dataset _strictly_ adheres to SITC codes.

The first dataset is created by defining a local dimension, `<local/dimension/product>`, with a local codelist, `<sitc-plus-extras>` which borrows codes from the SITC codelist but also includes bespoke codes.
The second dataset is created but adopts the family level sitc dimension, `<family/dimension/sitc>`, and sitc codelist directly.
 
I do not think the following can hold:
```
<local/dimension/product> rdfs:subPropertyOf <family/dimension/sitc>
``` 

This is because the codelist of the local dimension contains bespoke codes not included in the codelist at the family level.

```
<obs> <local/dimension/product> <bespoke-code> .
# Would imply that
<obs> <family/dimension/sitc> <bespoke-code> .
# But the second statement is invalid because <bespoke-code> is not part of the <sitc> concept scheme.
```

The Trade search makes use of `rdfs:subPropertyOf` relationships to guide the user to dimensions which are similar in nature.

Both dimensions are similar in that they represent products. 

The proposal is, as a temporary but logically valid way of producing a functional search, we create a number of "broad" dimensions such as "product" to act as parent properties of any dimensions which remotely relate to the concept of "product". These dimensions will have no codelists specified, and `rdfs:range` set to `rdfs:Resource` so as to keep them as "broad" and not introduce logical inconsistencies.

We recognise these dimensions serve no purpose other than producing an umbrella resource upon which to tie multiple, similar dimensions together in a way which allows the search to function. These dimensions would not be used in any dataset definition, they would only be used as part of `rdfs:subPropertyOf` relationships to "group" similar dimensions together.


```ttl
<local/dimension/product> a qb:DimensionProperty ;
    qb:codeList <sitc-plus-extras> ;
[...]

PR #229, opened 2021-04-06
Allow specifying only parent dimension

rossbowen:
See https://github.com/GSS-Cogs/gss-utils/issues/224.

For local dimensions with local codelists, we want `CSVWMapping()` to generate default URIs but also to have the ability for the user to specify a parent URI as part of the `info.json`.

For example,
```json
{
    "Region": {
        "parent": "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod"
    }
}
```

Should produce `.ttl` of the form:
```ttl
      <#dimension/region> a qb:DimensionProperty ;
          qb:codeList <#scheme/region> ;
          rdfs:subPropertyOf <http://purl.org/linked-data/sdmx/2009/dimension#refPeriod> .
```


ajtucker (approved):
Looks good! We should check and update the JSON schema.

PR #231, opened 2021-04-07
Alter how Graph URI is decided for datasets generated by gss-utils

robons:
https://github.com/GSS-Cogs/pmd-jenkins-library/issues/66

Jenkins will be altered to pull the Graph URI from the explicit `sd:NamedGraph` node instead of simply using the #dataset URI.

### DIFF
--- a/gssutils/csvw/mapping.py
+++ b/gssutils/csvw/mapping.py
@@ -397,7 +397,13 @@ class CSVWMapping:
         )
         self._validate()
 
-        containing_graph_uri = self._containing_graph_uri or self.join_dataset_uri("#graph")
+        if self._containing_graph_uri is None:
+            print("WARNING: _containing_graph_uri is unset. Imputing graph URI from context.")
+            containing_graph_uri = self._dataset_uri.replace("gss-data.org.uk/data/gss_data",
+                                                             "gss-data.org.uk/graph/gss_data")
+        else:
+            containing_graph_uri = self._containing_graph_uri
+
         csvw_structure = {
             "@context": ["http://www.w3.org/ns/csvw", {"@language": "en"}],
             "tables": self._as_tables(),
--- a/gssutils/csvw/mapping.py
+++ b/gssutils/csvw/mapping.py
@@ -44,6 +44,7 @@ class CSVWMapping:
         self._measureTemplate: Optional[URITemplate] = None
         self._measureTypes: Optional[List[str]] = None
         self._accretive_upload: bool = False
+        self._containing_graph_uri: Optional[URI] = None
 
     @staticmethod
     def namify(column_header: str):
@@ -110,6 +111,9 @@ class CSVWMapping:
         if self._foreign_keys is None:
             self._foreign_keys = []
         self._foreign_keys.append(foreign_key)
+
+    def set_containing_graph_uri(self, uri: URI):
+        self._containing_graph_uri = uri
 
     def set_dataset_uri(self, uri: URI, dataset_root_uri: Optional[URI] = None):
         f"""
@@ -392,15 +396,25 @@ class CSVWMapping:
             valueUrl=URI("qb:Observation")
         )
         self._validate()
+
+        containing_graph_uri = self._containing_graph_uri or self.join_dataset_uri("#graph")
         csvw_structure = {
             "@context": ["http://www.w3.org/ns/csvw", {"@language": "en"}],
             "tables": self._as_tables(),
-            "@id": self.join_dataset_uri("#tables")
+            "@id": containing_graph_uri,
+            # sd:NamedGraph => https://www.w3.org/TR/sparql11-service-description/#sd-NamedGraph
+            "rdf:type": {
+                "@id": "sd:NamedGraph"
+            },
+            "sd:name": {
+                "@id": containing_graph_uri
+            }
         }
 
         if not self._accretive_upload:
             # Don't want to upload DSD twice where we're just adding new data to existing data.
-            csvw_structure["prov:hadDerivation"] = DataSet(
+            # void:rootResource => https://www.w3.org/TR/void/#root-resource
+            csvw_structure["void:rootResource"] = DataSet(
                 at_id=self.join_dataset_uri('#dataset'),
                 qb_structure=DSD(
                     at_id=self.join_dataset_uri('#structure'),
--- a/gssutils/transform/cubes.py
+++ b/gssutils/transform/cubes.py
@@ -107,6 +107,7 @@ class Cube:
 
         map_obj.set_csv(destination_folder / f'{pathified_title}.csv')
         map_obj.set_dataset_uri(urljoin(self.scraper._base_uri, f'data/{self.scraper._dataset_id}'))
+        map_obj.set_containing_graph_uri(self.scraper.dataset.graph)
 
         return map_obj
 
--- a/gssutils/transform/cubes.py
+++ b/gssutils/transform/cubes.py
@@ -107,7 +107,7 @@ class Cube:
 
         map_obj.set_csv(destination_folder / f'{pathified_title}.csv')
         map_obj.set_dataset_uri(urljoin(self.scraper._base_uri, f'data/{self.scraper._dataset_id}'))
-        map_obj.set_containing_graph_uri(self.scraper.dataset.graph)
+        map_obj.set_containing_graph_uri(self.scraper.dataset.pmdcatGraph)
 
         return map_obj
 
--- a/gssutils/transform/cubes.py
+++ b/gssutils/transform/cubes.py
@@ -5,6 +5,7 @@ import copy
 
 from pathlib import Path
 from urllib.parse import urljoin
+from typing import Optional
 
 from gssutils.csvw.mapping import CSVWMapping
 from gssutils.utils import pathify
@@ -35,11 +36,12 @@ class Cubes:
             logging.warning("The passing of job_name= has been depreciated and no longer does anything, please"
                             "remove this keyword argument")
 
-    def add_cube(self, scraper, dataframe, title, graph=None, info_json_dict=None):
+    def add_cube(self, scraper, dataframe, title, graph=None, info_json_dict=None, override_containing_graph=None):
         """
         Add a single datacube to the cubes class.
         """
-        self.cubes.append(Cube(self.base_uri, scraper, dataframe, title, graph, info_json_dict))
+        self.cubes.append(Cube(self.base_uri, scraper, dataframe, title, graph, info_json_dict,
+                               override_containing_graph))
 
     def output_all(self):
         """
@@ -82,8 +84,10 @@ class Cube:
     """
     A class to encapsulate the dataframe and associated metadata that constitutes a single datacube
     """
+    override_containing_graph_uri: Optional[str]
 
-    def __init__(self, base_uri, scraper, dataframe, title, graph, info_json_dict):
+    def __init__(self, base_uri, scraper, dataframe, title, graph, info_json_dict,
+                 override_containing_graph_uri: Optional[str] = None):
 
         self.scraper = scraper  # note - the metadata of a scrape, not the actual data source
         self.dataframe = dataframe
@@ -91,6 +95,7 @@ class Cube:
         self.scraper.set_base_uri(base_uri)
         self.graph = graph
         self.info_json_dict = copy.deepcopy(info_json_dict)  # don't copy a pointer, snapshot a thing
+        self.override_containing_graph_uri = override_containing_graph_uri
 
     def _instantiate_map(self, destination_folder, pathified_title, info_json):
         """
@@ -107,7 +112,11 @@ class Cube:
 
         map_obj.set_csv(destination_folder / f'{pathified_title}.csv')
         map_obj.set_dataset_uri(urljoin(self.scraper._base_uri, f'data/{self.scraper._dataset_id}'))
-        map_obj.set_containing_graph_uri(self.scraper.dataset.pmdcatGraph)
+
+        if self.override_containing_graph_uri:
+            map_obj.set_containing_graph_uri(self.override_containing_graph_uri)
+        else:
+            map_obj.set_containing_graph_uri(self.scraper.dataset.pmdcatGraph)
 
         return map_obj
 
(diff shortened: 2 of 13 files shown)


gssutils/csvw/mapping.py
```diff
@@ -392,15 +396,25 @@ def get_value_uri_template_for_col(column_def: object, column_name: str) -> URI:
             valueUrl=URI("qb:Observation")
         )
         self._validate()
+
+        containing_graph_uri = self._containing_graph_uri or self.join_dataset_uri("#graph")
```
rossbowen, line 400: Will this coin:
`http://gss-data.org.uk/data/somedataset#graph`? 
It's different to what's used elsewhere.

rossbowen (commented):
Comments below, also need to think about how this interacts with metadata derived for PMD:
https://github.com/GSS-Cogs/gss-utils/blob/c740d469299ec71a753118d8bbd994e52a60a78d/gssutils/scrape.py#L383

gssutils/csvw/mapping.py
robons: Indeed, it will. We're currently using `#dataset` as the graph URI, but this will help us distinguish the graph from the dataset itself.

ajtucker (commented):
Can you at least have a think about how to write the tests such that they’re not so closely tied to the implementation?

robons:
> Can you at least have a think about how to write the tests such that they’re not so closely tied to the implementation?

Any suggestions as to how? The tests need to ensure that the RDF meets expectations. Is there anyother way to test that what we output is correct?

gssutils/csvw/mapping.py
rossbowen: I thought that `/graph/` takes the place of `/data/`, e.g. `http://gss-data.org.uk/graph/somedataset`.

robons: Oh yes, you're right. The graph URI comes from the TRIG file. I should probably alter this to pick up the graph URI from wherever in gss-utils generates that trig file.

rossbowen:
Thanks for the meeting this morning on this one. Would be helpful if you could check my understanding.

So our motivation is that, I have some dataframe `df` which is a bit big to handle in PMD. We want to partition the data into multiple RDF graphs.

This approach will add the following RDF relationships to _all_ `.ttl` outputs.

```ttl
@prefix sd: <http://www.w3.org/ns/sparql-service-description#>.
@prefix void: <http://rdfs.org/ns/void#>.

<http://gss-data.org.uk/graph/gss_data/trade/example> a sd:NamedGraph ;
    sd:name <http://gss-data.org.uk/graph/gss_data/trade/example> ;
    void:rootResource <http://gss-data.org.uk/data/gss_data/trade/example#dataset> .
```

It does this by adding several triples to the `-metadata.json` file which is created by the `Cubes()` class.

The dataset `df` is wrangled by the engineer into some sub-frames, `df_1`, `df_2` etc. and then our code will broadly look of the form:

```python
scraper = Scraper(info)
cubes = Cubes(info)
cube = Cubes()

cube.add_cube(scraper, df_1, some_graph_id_1, some_dataset_id_1)
cube.add_cube(scraper, df_2, some_graph_id_2, some_dataset_id_2)
cube.add_cube(scraper, df_3, some_graph_id_3, some_dataset_id_3)
```

This results in three `-metadata.json` files being produced, but only one `.trig` file containing the "umbrella" dataset metadata which applies across all chunks.

The `.trig` file currently specifies a singular graph for the dataset resource, and we need to check with Swirrl whether this would need to be represented as specifying multiple graphs, something like this:

```ttl
<http://gss-data.org.uk/data/gss_data/covid-19/example-catalog-entry> a pmdcat:Dataset ;
    pmdcat:datasetContents <http://gss-data.org.uk/data/gss_data/covid-19/example#dataset> ;
    pmdcat:graph ns1:example-01, ns1:example-02, ns1:example-03 .
```
---------

I suppose my main commentary is that we're polluting our `-metadata.json` files with metadata which is very specific to our usage of the data. I assume Jenkins will seek this data to make decisions about where to put the data in PMD etc. which results in some tight coupling. I'd be interested to know if there are "post-processing" approaches which keep the graph detail out of our artifacts, but that may go against conversations we've had elsewhere.

If we do pursue this approach, accepting that it's a temporary measure until we factor out the PMD specific detail from gssutils, I would perhaps consider whether this should be controlled by an argument somewhere, rather than pushing this approach across all datasets.

robons:
> Thanks for the meeting this morning on this one. Would be helpful if you could check my understanding.
> 
> So our motivation is that, I have some dataframe `df` which is a bit big to handle in PMD. We want to partition the data into multiple RDF graphs.
> 
> This approach will add the following RDF relationships to _all_ `.ttl` outputs.
> 
> ```turtle
> @prefix sd: <http://www.w3.org/ns/sparql-service-description#>.
> @prefix void: <http://rdfs.org/ns/void#>.
> 
> <http://gss-data.org.uk/graph/gss_data/trade/example> a sd:NamedGraph ;
>     sd:name <http://gss-data.org.uk/graph/gss_data/trade/example> ;
>     void:rootResource <http://gss-data.org.uk/data/gss_data/trade/example#dataset> .
> ```
> 
> It does this by adding several triples to the `-metadata.json` file which is created by the `Cubes()` class.
> 

Correct so far.

> The dataset `df` is wrangled by the engineer into some sub-frames, `df_1`, `df_2` etc. and then our code will broadly look of the form:
> 
> ```python
> scraper = Scraper(info)
> cubes = Cubes(info)
> cube = Cubes()
> 
> cube.add_cube(scraper, df_1, some_graph_id_1, some_dataset_id_1)
> cube.add_cube(scraper, df_2, some_graph_id_2, some_dataset_id_2)
> cube.add_cube(scraper, df_3, some_graph_id_3, some_dataset_id_3)
> ```
> 
Kind of, perhaps @canwaf can confirm here.  

What we *currently* actually do is to create one CSV-W per pipeline run. Creating one chunk, converting to RDF and uploading to a clean graph already takes 3+ hours (without CSV-W validation being attempted... that takes 4+ hours). I don't believe that we want to start creating, converting and uploading multiple chunks in the same Jenkins pipeline until we're more confident that our approach is reliable.

> This results in three `-metadata.json` files being produced, but only one `.trig` file containing the "umbrella" dataset metadata which applies across all chunks.
> 
The `.trig` file is only generated on the first upload. The first upload is special in that it is the only time that codelists, the `qb:DataSet` declaration and the data structure are generated and uploaded.

Subesquent uploads do not generate a `.trig` file, nor do they declare the `qb:DataSet` nor the DSD (which is already defined by the first upload).

> The `.trig` file currently specifies a singular graph for the dataset resource, and we need to check with Swirrl whether this would need to be represented as specifying multiple graphs, something like this:
> 
> ```turtle
> <http://gss-data.org.uk/data/gss_data/trade/example-catalog-entry> a pmdcat:Dataset ;
>     pmdcat:datasetContents <http://gss-data.org.uk/data/gss_data/trade/example#dataset> ;
>     pmdcat:graph ns1:example-01, ns1:example-02, ns1:example-03 .
> ```
> 
We confirmed this on slack yesterday. We do need to generate a `?catalog-entry pmdcat:dataset <http://containing-graph-uri>` triple for every graph which contains DSD/slices/observations related to the dataset. PMD unions these graphs together when querying and displaying the data.

> I suppose my main commentary is that we're polluting our `-metadata.json` files with metadata which is very specific to our usage of the data. I assume Jenkins will seek this data to make decisions about where to put the data in PMD etc. which results in some tight coupling. I'd be interested to know if there are "post-processing" approaches which keep the graph detail out of our artifacts, but that may go against conversations we've had elsewhere.
> 
Yes, this is correct. We are polluting our CSV-Ws. I think a better way to approach this problem in future would be to require users to provide a partition key for their datasets which we could use to automatically define `qb:Slice`s. We could then partition all observations by these slices and automatically upload them to separate graphs which could re-use the slice's URI.

If the user specified that the `Year (period)` was the partition key, this would make it easy to write highly performant queries focusing on a particular year's data. It should also make the data easier to browse in future if and when PMD support some UI interface based on slices. It would also ensure that we're being sneaky about partitioning data without explicitly requiring users to define Graph URIs.

> If we do pursue this approach, accepting that it's a temporary measure until we factor out the PMD specific detail from `gssutils`, I would perhaps consider whether this should be controlled by an argument somewhere, rather than pushing this approach across all datasets.
This approach only adds a couple of triples. If we add an on/off switch in gss-utils to support turning this functionality off, this means that we have to provide two separate branches of the pmd-jenkins build pipeline to support extracting the graph URI from the `.trig` file (as we currently do) *and* supporting the new method of extracting the graph URI from the CSV metadata file. I think it'd be simplest just to roll forwards on this one, leave the technical debt, and then tidy it up once we're on the other side and are able to spend more time thinking about the best long-term approach.

> ### _Notes on post-processing_
> 
> Some rough thinking:
> 
> Given some large dataframe `df` and a single `df.csv-metadata.json` file (as in the standard case), could I not:
> 
>     * Partition `df` into an arbitrarily amount of `df_1.csv`, `df_2.csv`, `...` files,
> 
>     * Apply `csv2rdf -t df_x.csv -u df.csv-metadata.json` an arbitrary amount of times,
> 
>     * Adjust the API calls to PMD so that for each of the `.ttl` outputs, when the data is added, a different graph parameter is specified?

It then becomes easier to accidentally duplicate data. If we explicitly specify the graph URI in the CSV-W metadata file, it's easy to check for graphs with the same URI which already exist (and so to delete their data). 

We also currently define the dataset's graph URI in the `.trig` file. So it's simpler to roll-forward with the approach of the graph URI being controlled by gss-utils and then completely remove all knowledge of graph URIs from gss-utils at a later point in time. I don't think having two places (gss-utils and pmd-jenkins shared library) which coin graph URIs is a good approach - I'd rather keep them in one place for now. 

> 
> 
> This keeps all knowledge of the graph outside of the `-metadata.json` file. Most of the heavy lifting would be done outside of `gssutils`.

canwaf:
> > The dataset `df` is wrangled by the engineer into some sub-frames, `df_1`, `df_2` etc. and then our code will broadly look of the form:
> > ```python
> > scraper = Scraper(info)
> > cubes = Cubes(info)
> > cube = Cubes()
> > 
> > cube.add_cube(scraper, df_1, some_graph_id_1, some_dataset_id_1)
> > cube.add_cube(scraper, df_2, some_graph_id_2, some_dataset_id_2)
> > cube.add_cube(scraper, df_3, some_graph_id_3, some_dataset_id_3)
> > ```
> 
> Kind of, perhaps @canwaf can confirm here.
> 
> What we _currently_ actually do is to create one CSV-W per pipeline run. Creating one chunk, converting to RDF and uploading to a clean graph already takes 3+ hours (without CSV-W validation being attempted... that takes 4+ hours). I don't believe that we want to start creating, converting and uploading multiple chunks in the same Jenkins pipeline until we're more confident that our approach is reliable.
> 

You're both right, @rossbowen understanding is how we manage cubes in the `cubes()` class and not the specific implementation. In a big dataset as you point out, @robons, we intend to only manage one cube/run of the pipeline.

gssutils/transform/cubes.py
```diff
@@ -82,15 +84,18 @@ class Cube:
     """
     A class to encapsulate the dataframe and associated metadata that constitutes a single datacube
     """
+    override_containing_graph_uri: Optional[str]
 
-    def __init__(self, base_uri, scraper, dataframe, title, graph, info_json_dict):
+    def __init__(self, base_uri, scraper, dataframe, title, graph, info_json_dict,
+                 override_containing_graph_uri: Optional[str] = None):
```
mikeAdamss, line 90: you don't really need the keyword here, when you add the cube (line 39, passed onwards on 44) `override_containing_graph_uri` is already either populated or None.

it doesn't hurt either (and there's a case for including it if you forsee called `Cube` without `Cubes`) so I'm not much bothered, but if we keep it can we change the line 44 entry to `override_containing_graph= override_containing_graph` otherwise we're method overloading, which works but is little logically mucky.

gssutils/transform/cubes.py
```diff
@@ -82,15 +84,18 @@ class Cube:
     """
     A class to encapsulate the dataframe and associated metadata that constitutes a single datacube
     """
+    override_containing_graph_uri: Optional[str]
```
mikeAdamss, line 87: You shouldn't need this.

mikeAdamss (commented):
a few minor comments but as far as I'm following what we're doing this makes sense to me 👍

gssutils/transform/cubes.py
robons: I know, but it helps with intellisense.

rossbowen:
Thanks @robons, @canwaf.

So thinking a bit longer term, I think there is a shared understanding and vision that:
- The CSVW artifact would be free of any graph information or anything bespoke/specific to how the data is organised within PMD.
- There is some information in the `.trig` file we currently generate, in particular `dcat` triples, which probably _do_ belong in the CSVW.

So with that in mind, including graph information which is bespoke/specific to how the data is organised within PMD certainly feels like a step backward but does enable in the short term a way a method to partition a dataset in different graphs which, if nothing else, will start to get us thinking about a better approach to this problem.

To be clear about the debt we're introducing - there will be a change to the Jenkins pipeline to seek `?dataset sd:NamedGraph ?graph` triples from any generated CSVW, and those triples will be populated in the CSVW output generated by the `Cubes()` class. The user may set a `override_containing_graph_uri` argument when initialising a `Cube()`.

So long as we're consistent in our understanding that, even in the medium term, the CSVW should not and must not be the home for graph information then I accept that this is debt and propose that https://github.com/GSS-Cogs/pmd-jenkins-library/issues/15 become one of our priorities.

In the meantime I will approve (though you may want to have a look at Mike's comments above). @robons when you begin to tackle the changes to the Jenkins pipeline I'd ask if you raise a PR there and add some narrative about what parts of the pipeline still remain the responsibility of the `.trig` file, and what parts are now the responsibility of the graph triples being produced as part of the CSVW. Getting a high view of the dependencies of the pipeline and which of those dependencies we consider debt will be informative.

gssutils/transform/cubes.py
robons: I've removed it the default param from the `Cube` constructor.

robons:
Right, I managed to get it to run the tests and they pass. All feedback taken account of. Good to approve?

rossbowen (approved):

PR #232, opened 2021-04-07
Check that using Cubes() does ignore JOB_NAME

ajtucker:
as long as ID and family are in info.json


mikeAdamss (approved):
I've approved as the functionality what we're testing here makes sense, but am I right in saying all the valueUrls in the column mapping fixture are now wrong?

ajtucker:
> I've approved as the functionality what we're testing here makes sense, but are all the valueUrls in the column mapping fixture wrong now? (question only, I'm not 100% on that, it's getting hard to follow).

Good question! I don't know. This issue has turned into one that tries to understand the implications, so I guess we should explicitly test that assumption!
 
> As an aside, I'm starting to think we _really_ need to start generating our uri's in the same place.

Yes. Though it's tricky when we're trying to make sure we're doing what we want the publishers to do.

PR #237, opened 2021-04-19
Use environment variable sparql endpoint for PMD chunking

### LINKED ISSUE #236: Use environment variable sparql endpoint for PMD chunking
what it says on the tin

canwaf:
Fixes #236

### DIFF
--- a/gssutils/transform/download.py
+++ b/gssutils/transform/download.py
@@ -8,6 +8,7 @@ import pandas as pd
 import pyexcel
 import requests
 import xypath
+from os import environ
 from typing import Union, Dict, Optional
 
 from gssutils.metadata.base import Resource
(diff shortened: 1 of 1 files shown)


robons:
@rossbowen @ajtucker Do we have the ability to set environment variables like this on groups of builds? i.e. can the builds in [jenkins/beta.gss-data.org.uk](https://ci.floop.org.uk/job/GSS_data/job/beta.gss-data.org.uk/) easily be configured to all point at the beta site?

rossbowen:
@robons I showed @canwaf the approach used [here](https://github.com/GSS-Cogs/ref_period/blob/6a459f28ea32a6c0436d2adc61333581e2e11ffc/Jenkinsfile#L20) which I think could be used.

The beta pipelines have a json config associated with them which is different to other pipelines, and the config file contains the location of the respective SPARQL endpoint.

We might need to make an additional change to the Jenkinsfile to make sure the environment variable gets set - wonder if you would agree?

rossbowen (approved):

canwaf:
Thanks @rossbowen !

PR #242, opened 2021-04-25
Issue 239/govuk json response issues

mikeAdamss:
I can't consistently recreate this issue.

On Friday the tests  would pass and fail sporadically (same tests with `record_mode=all`). Literally the same code with no particular pattern.

the test monitoring has everything back to working this morning and I'm not getting the fails (though again, that could be random).
Screenshot 2021-04-25 at 11 03 52

so I _think_ this is on their end and it may be that it's been resolved but I'm not 100% and it's pretty difficult to ensure the absence of a random factor regardless.

so as a precaution this pr is (hopefully temporarily) wrapping said endpoints with exponential backoff for the json decode error. Not subtle, but I'm hoping this should mitigate this problem if it reoccurs at a bad time.


mikeAdamss:
this problem this was aimed at is still absent. Will leave this on the branch for a week or two, if it doesn't reoccur we can just close this.

mikeAdamss:
closing, was just a blip on their apis.

Issue #247, opened 2021-05-04
Suppress catalog and qb:DataStructureDefinition for multi-cube uploads


canwaf:
Due to the size of CSV-Ws for family-trade/issue9, we need multiple CSV-Ws generated and follow the multi-graph paradigm without accretive upload. This involves specifying at `Cube().add_cube()` a flag which suppresses the generation of CSV-W catalog and qb:DSD for additional 'chunks' beyond the primary graph (i.e. the secondary graphs).

- [ ] add flag to `add_cube()` method & pass through to CSVWMapper
- [ ] suppress DSD and catalog generation with on new flag being true
- [ ] Add behave test to ensure DSD and trig not output when requested

ajtucker:
The [CSV2RDF specification](https://www.w3.org/TR/csv2rdf/) offers up two modes of operation: minimal and standard.

[Swirrl's implementation](https://github.com/Swirrl/csv2rdf) adds another mode, annotated, which we're using to carry across the extra JSON-LD metadata in the `-metadata.json` file into the resulting triples.

We should be able to split out the metadata from the observations by taking advantage of this.

PR #248, opened 2021-05-04
Issue #247 - Suppress catalog and DSD metadata independently of accretive upload functionality

robons:
In order to add support for multi-graph datasets without having to use the accretive upload functionality, we need to be able to suppress the catalog metadata and DSD metadata from being output on a cube-by-cube basis.

The primary graph (file) should contain the DSD and have a complementary `.trig` file. The secondary graph files should not have either of these things (to avoid duplication).

This PR allows us to suppress these bits of metadata on a cube-by-cube basis. The decision on which cube contains the DSD and has the associated catalogue metadata will be decided in each transformation's main.py file.

### DIFF
--- a/features/csvw.feature
+++ b/features/csvw.feature
@@ -389,7 +389,7 @@ Scenario: Ensure column definition `value` field is respected when `parent` is n
           .
     """
 
- Scenario: Do not output Data Structure Definition or Trig when performing an accretive upload
+ Scenario: Do not output Data Structure Definition when performing an accretive upload
     Given a CSV file 'product-observations.csv'
     And a JSON map file 'mapping-info-accretive-test.json'
     And a dataset URI 'http://gss-data.org.uk/data/gss_data/trade/ons-uk-trade-in-goods-by-industry-country-and-commodity'
@@ -399,6 +399,16 @@ Scenario: Ensure column definition `value` field is respected when `parent` is n
     And the RDF should pass the Data Cube integrity constraints
     And the ask query 'dsd-components-exist.sparql' should return False
 
+ Scenario: Do not output Data Structure Definition when output suppressed
+    Given a CSV file 'product-observations.csv'
+    And a JSON map file 'mapping-info.json'
+    And a dataset URI 'http://gss-data.org.uk/data/gss_data/trade/ons-uk-trade-in-goods-by-industry-country-and-commodity'
+    And and the catalog and DSD metadata output should be suppressed
+    When I create a CSVW file from the mapping and CSV
+    Then the metadata is valid JSON-LD
+    And gsscogs/csv2rdf generates RDF
+    And the RDF should pass the Data Cube integrity constraints
+    And the ask query 'dsd-components-exist.sparql' should return False
 
 Scenario: A user wishes to specify a parent dimension for a locally defined dataset dimension
     Given a CSV file 'life-expectancy.csv'
--- a/features/cubes.feature
+++ b/features/cubes.feature
@@ -51,3 +51,16 @@ Feature: Creating cubes
 
     <http://containing-graph-uri> a sd:NamedGraph.
     """
+
+    Scenario: Suppressing catalog and DSD output should ensure trig file not generated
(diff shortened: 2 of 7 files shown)


canwaf (approved):

rossbowen:
What's the usage here - as a DE with a large dataframe held in some variable `df`, what do I now do?

robons:
> What's the usage here - as a DE with a large dataframe held in some variable `df`, what do I now do?

@canwaf is currently working on a build script we'll be adding into the Jenkins pipeline tests. This should be a good location to document how to use it. 

Essentially, imagine you're building up a large dataset, you have it in a data frame, you split it up into chunks and add each of them individually to the cubes class using `cubes.add_cube(...)`. When making your call to `cubes.add_cube(....)`, you'll let cubes generate the DSD and trig metadata for the **first** graph/file/cube and then tell it to suppress output of those for your following cubes.

This ensures that your primary graph has the DSD and trig associated with it, but the secondary graphs don't.

rossbowen:
Righto. So `add_cube` is a little inaccurate as a name as it's being called multiple times to form a single cube. When we remove this option later on, any pipelines using the feature will break too so I suppose we need to use sparingly.

I'm assuming no changes to the Jenkins build pipeline are needed, but you mention a test-case being added above so just want to double check that the Jenkins pipeline doesn't require a change.

robons:
> Righto. So `add_cube` is a little inaccurate as a name as it's being called multiple times to form a single cube. When we remove this option later on, any pipelines using the feature will break too so I suppose we need to use sparingly.
> 
Yes, I think there's a lot of overlap of what `cube` means, it can change in different context, but I agree it's misleading here. I believe `cubes.add_cube` is being used is all current transforms. It's the recommended way foreward. If what you're thinking of doing is removing or renaming `add_cube`, I'd speak with @mikeAdamss.

> I'm assuming no changes to the Jenkins build pipeline are needed, but you mention a test-case being added above so just want to double check that the Jenkins pipeline doesn't require a change.

Indeed, I'm assuming that no changes to the Jenkins pipeline are necessary, but I'm not 100% confident so we're adding this test case to fix any small problems we find. Since the multi-graph functionality has only been tested with accretive uploads, there could be some problems lurking in the dark when we allow them without it being an accretive upload.

rossbowen:
I think it's probably an indication of `add_cube` and `Cube` venturing away from the original intention rather than me suggesting a rename!

Feeling nervous around all this, as you know. Would be good in the next sprint to agree some thinking around https://github.com/GSS-Cogs/pmd-jenkins-library/issues/15 if you'd planned to push on any more of this sorts of stuff.

mikeAdamss:
catching up (things to fix today), yes `add_cube()` is arguably misnamed now (has been for a while) have no issues with renaming it, or better yet just wrap creating the `Cube` (or thing we currently call Cube) in a different constructor (`add_cube_slice()` or somesuch), if indeed there are any non loaded terms left.

that's not me waving a "gogogo" flag btw, because we can doesn't mean we should, this feels like it all needs a think.

mikeAdamss:
somethings itching my brain about this pr. Are there any scenarios where you'd `override_containing_graph ` and _not_ `suppress_catalog_and_dsd_output` ?

Likewise, would you ever `suppress_catalog_and_dsd_output` and _not_ `override_containing_graph`? (this one especially)

As in, is the reason we're suppressing catalog and dsd _because_ we're overriding the containing graph?

Just sharing my thinking, Trying to figure out if we're saying the same thing twice here.

robons:
> somethings itching my brain about this pr. Are there any scenarios where you'd `override_containing_graph ` and _not_ `suppress_catalog_and_dsd_output` ?
> 
> Likewise, would you ever `suppress_catalog_and_dsd_output` and _not_ `override_containing_graph`? (this one especially)
> 
> As in, is the reason we're suppressing catalog and dsd _because_ we're overriding the containing graph?
> 
> Just sharing my thinking, Trying to figure out if we're saying the same thing twice here.

I'm trying not to couple the two things together. I suspect we'll want to be able to vary the graph URI in future when publishers start using their own URIs.

PR #249, opened 2021-05-04
Re-instate some foreign key constraints

ajtucker:
Local dimensions with no codelist specified should validate against the codes declared in a local `codelists` directory if it exists.


gssutils/csvw/mapping.py
```diff
@@ -203,7 +207,24 @@ def get_value_uri_template_for_col(column_def: object, column_name: str) -> URI:
...
             return get_conventional_local_codelist_concept_uri_template(column_name)
 
-        # Look to see whether the measure type has its own column
+        def add_local_codelist(name: str):
+            if self._codelist_base is not None:
+                codelist_csv = (self._codelist_base / pathify(name)).with_suffix('.csv')
+                codelist_relative_uri = URI(codelist_csv)
+                self._external_tables.append(Table(
+                    url=codelist_relative_uri,
+                    tableSchema=URI("https://gss-cogs.github.io/family-schemas/codelist-schema.json"),
```
robons, line 216: I wouldn't agree that *all* codelists use the same schema. There are certainly some which vary. I reckon we should probably specify that the schema is at `str(codelist_csv) + '-metadata.json'`

features/csvw.feature
```diff
@@ -466,4 +467,49 @@ Scenario: Manually Overriding CSV-W's graph URI works.
         a sd:NamedGraph;
         sd:name <http://gss-data.org.uk/graph2/gss_data/trade/ons-uk-trade-in-goods-by-industry-country-and-commodity>;
         void:rootResource <http://gss-data.org.uk/data/gss_data/trade/ons-uk-trade-in-goods-by-industry-country-and-commodity#dataset>.
-    """
\ No newline at end of file
+    """
+
+  Scenario: Validate local codelists
```
robons, line 472: I would say it'd be a good idea to have a test which ensures that validation fails when one of the values doesn't match the expected foreign key constraint. That is what's being implemented here.

robons (commented):
For anyone else who's going to review this, I'd suggest starting at the bottom of the PR and working your way up - this is the way that made the most sense to me.

gssutils/csvw/mapping.py
ajtucker: While true that not all codelists have the same schema, I'm going to leave this as the default and we can think about overriding as needed. Feel free to make a separate issue. Note that the full tabular metadata JSON `-metadata.json` is different to the table schema JSON (tabular metadata contains table schemas).

Note also that I've updated the common schema with an optional `Description` column, so this schema now validates against all the local codelist CSV files we have in trade. In the end, this PR is about validating that the foreign keys are correct, not about validating the codelists themselves.

features/csvw.feature
ajtucker: Good catch. See b7b3c5d91fc2f003ae8c76d393697c51e542c20b

robons (approved):

robons:
I believe this PR is causing issues in dataset builds where there Date/Time concept schemes are created on-the-fly in Jenkins.

When running the integration tests in [pmd-jenkins-pipeline](github.com/GSS-Cogs/pmd-jenkins-library/), I receive the following error:
```
+ csvlint --no-verbose -s out/ons-uk-total-trade.csv-metadata.json
/var/folders/0p/5mskt39n2l9_bw46x82hm6d80000gq/T/jkh15740670034418153381/workspace/project/datasets/TestJob/codelists/period.csv not found
```

I believe this is a result of the CSV-W being configured so as to expect that locally defined date/time code lists have a CSV file containing the definitions. Seeing as these are generally generated on-the-fly in Jenkins, the codelist isn't available at the point in time where we currently ensure that the CSV-W output is valid. 

One potential fix is to alter the pmd-jenkins-library build pipeline so that it attempts to validate dataset CSV-Ws after we've generated the date/time codelists. What do people think?

ajtucker:
Does #257 cover this?

robons:
Yes, issue #257 would fix this problem.

PR #252, opened 2021-05-10
Issue 243/cubes cmd pmd outputs

mikeAdamss:
This PR is for the pupose of upgrading the cubes classes to allow them to create ouputs to satisfy multiple data platforms (PMD and CMD) via a single dataset transform.

To do this, I had to decouple us from the graph-based and pmd-based assumptions in the cube(s) classes, so this ties directly into some of our thoughts on how to approach refactoring (at the very least, as an example of decoupling via python abstract classes).

# User Side Changes

Imagine this simple transform for a PMD4 dataset

```python
scraper = Scraper(seed="info.json")
cubes = Cubes("info.json")

for tab in scraper.distribution(latest=True):
    # databaker stuff

    # SOME PMD SPECIFIC DATASET CHANGES

    cubes.addcube(scraper, df. tab.name)

cubes.output_all()
```

The pattern to make it a PMD4 and CMD transform would be

```python
scraper = Scraper(seed="info.json")
cubes = Cubes("info.json")

def pmd_func(df):
    # SOME PMD4 SPECIFIC DATASET CHANGES
    return df

def cmd_func(df):
    # SOME CMD SPECIFIC DATASET CHANGES
    return df

for tab in scraper.distribution(latest=True):
    # databaker stuff
    cubes.addcube(scraper, df, tab.name, writers = {PMD4Writer: pmd_func, CMDWriter: cmd_func})

cubes.output_all(writers = [PMD4Writer, CMDWriter])
```
So the idea is we do a single transform and pivot to a point in common and add it to the cube - then - make any platform specific tweaks via the attached functions.

This works almost entirely of the column mapping that we're already using, the only info.json addition is where a CMD dimension also has an associated labels column (not a thing we do on pmd4 at the moment) you specify it as follows. This is purely so we can correctly sequence the additional columns to satisfy the implicit cmd column schema.

For example if you had an "Industry Section" dimension in the column mapping, you'd specify it's associated lable column as follow:

```json
"transform":
        "cmd": {
                "associated_notations": {
                    "Industry Section": "Industry Section Label"
                }
            }
```

There's quite a few variations of how you can attach the writers and formaters (to cover the variety of potential use cases) so I'll explain that a bit below.

# Why are we passing the writers in twice?

You wouldn't, it's just a simple example of the api. You can pass in the writers and the formaters at either/both the Cubes() and Cube() level, so all of the following are valid.

Declare writers globally (so for all cubes), this is our common use case.
```python
for tab in scraper.distribution(latest=True):
    # databaker stuff
    cubes.addcube(scraper, df, tab.name)

cubes.output_all(writers = {PMD4Writer: pmd_func, CMDWriter: cmd_func})
```

Explicitly declare writers and formaters per cube
```python
for tab in scraper.distribution(latest=True):
    # databaker stuff
    cubes.addcube(scraper, df, tab.name, writers = {PMD4Writer: pmd_func, CMDWriter: cmd_func})

cubes.output_all()
```

Declare globally, but an extra formater for one specific cube, so:
```python
cubes.addcube(scraper, df, "cube 1")

# Extra formating for the CMD output of this _specific_ cube
cubes.addcube(scraper, df, "cube 2", writers = {CMDWriter: cmd_func2})

cubes.output_all(writers = {PMD4Writer: pmd_func, CMDWriter: cmd_func})
```
* _CMD output for cube1 uses cmd_func_
* _CMD output for cube2 uses cmd_func, cmd_func2_
* _PMD output for cube1 uses pmd_func_
* _PMD output for cube2 uses pmd_func_

Declare globally, but override to _only_ use Cube() level writer for an edge case
```python
cubes.addcube(scraper, df, "cube 1")

# Only write this cube to cmd and only formated with cmd_func2
cubes.addcube(scraper, df, "cube 2", writers = {CMDWriter: cmd_func2}, use_global_writers = False)

cubes.output_all(writers = {PMD4Writer: pmd_func, CMDWriter: cmd_func})
```
* _CMD output for cube1 uses cmd_func_
* _CMD output for cube2 uses cmd_func2_
* _PMD output for cube1 uses pmd_func_
* _PMD output for cube2 DOES NOT EXIST_

You can also pass in writers without formaters at either level as either a singleton of list if no formater declaration is needed, as follows:

```python
cubes.output_all(writers = [PMD4Writer, CMDWriter])
# or
cubes.output_all(writers = PMD4Writer)
```

_Note: if no writer is specified the default is `PMD4Writer` which is how we maintain backwards compatibility._

# Sub Classing Cube

I haven't dug into this too much, but I've added a tweak to allow extending `Cube()` for edge cases if we need to. so `add_cube` can now do one of two things.

* Take the required arguments to create a simple cube.
* Take an **already instantiated** cube or subclass of.

The latter should be the edge case, but it gives us the ability to extend functionality where needed, for example:

1.) Create a subclass of Cube to do something that Cube() doesn't
```python
class FancyCube(Cube):
    def clever_extra_thing(self):
        # Something clever
```

2.) Create a driver for this edge case (if needed).

3.) Instanitate the cube (with whatever params your driver requires) then pass it in, eg
```python
my_fancy_cube = FancyCube(scraper, df_out, "title is this", writers=FancyCubeDriver)

my_fancy_cube.clever_extra_thing()

cubes.add_cube(my_fancy_cube)
cubes.output_all()
```

Being able to extend the input class and define new drivers as needed _should_ allow us to meet edge cases without making any breaking changes to the api.

# Decoupling

The cube classes in this pr are decoupled from any assumptions about pmd4 and graphs.

Now the Cube classes don't write the outputs, they must manage the "writers".

* There is now an abstract class known as CubeWriter
* There is an implementation of this called PMD4Writer that does the necessaries to create PMD4 inputs.
* There is an implementation of this called CMDWriter that does the necessaries to create CMD inputs.

The trick is that while assumptions still exist, they only exist in the correct places. For example: if you're writing using the PMD4Writer which is a class that generates a trig from metadata, it's reasonable to assume its metadata object has a `.generate_trig()` method.

But for another writer, the metadata doesn't need to be a scraper nor does it need that method (for that matter, you'd don't need to pass in a metadata object at all).

The idea is the cubes classes are now agnostic as to the implementation, they hold a selection of the variables passed to them (all optional keywords), attach the formaters and run the writers - but they don't care how those writers are implemented.

It also means we can drop in a new writer or extend `Cube()` whenever we want to without being beholden to old assumptions or locking ourselves into new ones. It's hard to get the point accross without getting into the abstract, but basically we're free to do a _lot_ now if we need to.

# note

I've taken care to do what I can with type hints, docstrings, comments, tests etc but this is a big, complex change with a lot of metaprogramming involved, there's no getting away from that. When this hits pr I would personally look at the cubes.feature, cubes,py and the writers as they are, rather than trying to track the changes.

You could pull cubes out of gss utils after this is in if you wanted to, but we probably _shouldn't_ (at least not until we sort out ci) just to avoid the pain of manually re-releasing it every time we unpick another component part of gssutils.

There _is_ an assumption of having a column mapping and info.json here, can't _not_ make that assumption until/unless we decouple the input part of the transforms as well.

Just to reiterate after the wall of dev-speak, this is 100% backwards compatible with the existing pipelines.


robons:
> "associated_notations": "Industry Section Label"

I don't believe adding the anything CMD-specific to the existing column definitions in the info.json is a good way going forward. I think we're uncertain exactly where we're going with the tooling, but it's quite possible that we'll end up wanting some external publishers to specify their own info.json files. To this end, I think it'd be better to create a separate map of column name => CMD associated notation inside the info.json file.

gssutils/transform/writers/cmd.py
```diff
@@ -0,0 +1,154 @@
...
+                ' a "Value" column via your column mapping.')
+
+        marker_column = [x for x in self.cube.observation_data if x in ["Markers", "Marker"]]
+        attribute_columns = {k:v for (k, v) in column_map.items() if "attribute" in v}
+
+        for unwanted_col in attribute_columns:
+            column_map.pop(unwanted_col)
+        column_map.pop("Value")
+
+        assert "Value" in self.cube.observation_data, (f'Column map is specifying a "Value" column '
```
robons, line 95: Do this work with multi-measure datasets - or is there an implicit requirement that it's a single-measure dataset and the value is in a column named `Value`?

gssutils/transform/writers/pmd4.py
```diff
@@ -0,0 +1,206 @@
...
+    def output_metadata(self):
+        """
+        Output the encapsulated metadata to the required format and
+        to the required place
+        """
+
+        # True, if all the statements are
+        if "load" in self.info_json:
+            is_accretive_upload = all([
+                not self.info_json,
```
robons, line 147: Not sure why we're expecting `self.info_json` to by None/False here. We've just checked that "load" is in `self.info_json`.

robons: Don't worry, I just wanted to understand how you'd done some stuff, so had a bit of a delve. Thought I'd highlight what I'd noticed incase I forgot.

gssutils/transform/writers/cmd.py
mikeAdamss: thats a great point, yes I was assuming a value column yeah, is there an obivous point in the column map that indicated the value column(s) when they not just called value?

robons: I'd look for an object containing both `unit` and `measure` (`datatype` is technically optional IIRC)

image

gssutils/transform/cubes/cubes.py
```diff
@@ -0,0 +1,287 @@
...
+        write_out_paths = {}
+        for writer in writers:
+            out_path: Path = writer.get_out_path()
+            if out_path in write_out_paths:
+                write_out_paths[out_path].append(writer)
+            else:
+                write_out_paths[out_path] = [writer]
+        for out_path, writers in write_out_paths.items():
+            if len(writers) > 1:
+                logging.warning(f'The out path {out_path} is being used by multiple writers,'
```
canwaf, line 230: Should this just be a warning? I'd stop the output here, since the PMD and CMD outputs would never be the same.

canwaf (approved):
This is a gargantian effort and takes how we want to start to develop in IDPD to the next level. I have no required changes, and if it is merged as is I'll absolutely have a play on getting it into some of my pipelines to see how it will best work.

In the CMD writer implementation of the Writer abstract class it is assumed that supressing DSD, catalog, and graph implementations which are used in the quad nature of stardog/Swirrl's PMD can be used to support chunked/accretive datasets. I think that functionalty needs to be confirmed. If CMD can't support that, it should be hidden behind a critical error for now. (Though it wouldn't be likely that we'd use these methods unintentionally.)

I was thinking about how to move data between the the Tidying and the (potentially) two or more writers. I've used SQLite3 to reduce memory footprint when pivoting (i.e. generating two views of the same sets of observations using group by with sum). This would also speed up large datasets' sorting, and projecting (which the cmdwriter does). It would be good to get your thoughts on using a common source that isn't necessarily in memory as the source of data, and whether that would fit in your abstract writer implementation.

Regardless: good to go.

gssutils/transform/cubes/writers/abstract.py
```diff
@@ -0,0 +1,161 @@
...
+        # this (for example, to access the data without the write for tests)
+        self.operational_sequence = (
+            self.set_inputs,
+            self.format_metadata,
+            self.validate_metadata,
+            self.format_data,
+            self.dynamic_format_data,
+            self.validate_data,
+            self.output_data,
+            self.output_metadata,
```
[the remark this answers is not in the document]
robons: The metadata is generated by iterating through the columns defined in the CSV. It does things like setting the column order in the metadata document (it's a CSV-W spec requirement that the columns are defined in the same order as they are found in the CSV). We also need to pick up columns defined in the CSV aren't defined in the info.json - these still need to be defined in the CSV-W metadata.

canwaf:
Junk comment, please ignore.

mikeAdamss:
This ones a little uncertain for now.

This dual load idea cant be actioned without the website dev team and even if they agreed with the approach (undecided) it wont get anywhere their roadmap until next year at which point they'll be able to ingest csvw directly (rendering a CMDWriter kinda pointless).

We probably **do** want the writer abstraction (its a useful thing in of itself) but I may need to remove the CMDWriter and address the pr comments before merging.

However, it's possible senior management (i.e Darren) might disagree with my assessment. So this is on hold until we get a nod to back burner it and strip out the CMDWriter from this pr.

mikeAdamss:
closing pr, cant see us coming back to this (will keep the branch for a bit as precaution).

PR #254, opened 2021-05-12
temp pin csv2rdf to get things working again

mikeAdamss:
if we need to only, don't just merge because it passes please.


mikeAdamss:
closing, Alex did it without the pin.

Issue #258, opened 2021-05-13
Local dimension codelists need `{+var_name}`


ajtucker:
A recent update to CSV2RDF has fixed an issue where we're not quite using the CSV-W specification correctly, so we need to chase out some bugs.

In HMRC RTS Small Area for example, we use '/' characters for codes, see https://github.com/GSS-Cogs/family-trade/blob/master/datasets/HMRC-RTS-Small-area/codelists/nuts-geography.csv. The CSV-W specification uses URI Templates ([RFC 6570](https://www.rfc-editor.org/info/rfc6570)) which will normally encode any reserved characters like `/` so that `nuts1/UKH` gets encoded as `nuts1%2FUKH` for instance. The way to avoid this percent-encoding is to use the `+` character as a value modifier, so instead of `{column_name}`, use `{+column_name}`.

This will need to be used for local dimensions/codelists and also for the observation URI itself.

ajtucker:
Created an [issue branch](/GSS-Cogs/gss-utils/tree/issue-258/local-dim-plus) and added a failing test.

Note that the `"aboutUrl"` template may also need `+` before the variables in order that the coordinates of the observation aren't unnecessarily encoded if they have reserved chars in them.

Note also that the space character will still be, and should be, percent encoded if it is used as a code.

PR #262, opened 2021-05-20
codelist-manager- Stop URL templates being escaped by CSV2RDF

robons:
Issue #258 - Adding upgrade to codelist-manager so that it ensures that all codelist valueUrl/propertyUrl/aboutUrl properties do not escape the value found inside the notation/parent_nodation columns.

### DIFF
--- a/features/codelist-manager.feature
+++ b/features/codelist-manager.feature
@@ -415,7 +415,7 @@ Scenario: Correct the "@id" when it's "#table" and we have skos:inScheme defined
       }
       """
 
-Scenario: Escape the notation URIS column templates where they haven't been escaped.
+Scenario: Correct escaped notation/parent_notation column URL templates.
     Given We have a metadata file of the form
       """
       {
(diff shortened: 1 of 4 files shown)


ajtucker:
Looks grand, thanks.

I can see the new scenario checks that the `+`s are in there, but does it check with some input like `st/uff`?

robons:
> Looks grand, thanks.
> 
> I can see the new scenario checks that the `+`s are in there, but does it check with some input like `st/uff`?

It's only looking at the metadata json file, so it doesn't check the actual values from the CSV file at all.

rossbowen (approved):

ajtucker:
> It's only looking at the metadata json file, so it doesn't check the actual values from the CSV file at all.

Ok. Is there a quick way to write a failing test that is then fixed by this? I.e. the test should check that the thing that wasn't working properly now works properly.

robons:
> Ok. Is there a quick way to write a failing test that is then fixed by this? I.e. the test should check that the thing that wasn't working properly now works properly.

I've already added a test ensuring that the output from the codelist-manager tool is correct. That's a new test I've added in this PR. I don't really want to be testing CSV2RDF as well here.

PR #264, opened 2021-05-20
Canwaf/issue263 return type binding now for as_pandas() and as_databaker()

canwaf:
This will help with python linting, type hints, and using pylance or pywrite.


gssutils/transform/download.py
```diff
@@ -43,10 +43,10 @@ def open(self):
         stream.decode_content = True
         return stream
 
-    def as_databaker(self, **kwargs):
+    def as_databaker(self, **kwargs) -> Union[List[xypath.Table], xypath.Table]:
```
robons, line 46: Does `_get_simple_databaker_tabs` ever return a single `xypath.Table` outside of a list?

canwaf: Without delving any deeper, I didn't want to write it so that it couldn't. I believe it returns a list of one.

robons: I believe it's only returning lists, so I'd leave it as `List[xypath.Table]` without the union to remove ambiguity.

Both of the branches in `_get_simple_databaker_tabs` force iterables -> lists before returning their values, so it's safe to specify that it's just returning a list.

canwaf: Union removed.

robons (approved):

PR #265, opened 2021-05-21
publised_location from environment variable

### LINKED ISSUE #245: publised_location from environment variable
Kludge of `info.json` for published_location (i.e. SPARQL endpoint) should be removed from gss-utils chunking, and associated tests as this is now coming from environment variables.

canwaf:
Fixes #245
* Just removes the info.json key/value pairs, as the default value fills in the blank


ajtucker (approved):
Looks fine. Published is with an `s`.

PR #270, opened 2021-06-03
allow mapping of attribute with or without value & scenario's to chec…

Shannon95:
…k outcomes are valid.


features/fixtures/business-characteristics-data-tables.csv
```diff
@@ -0,0 +1,2101 @@
+Value,Flow,Year,Country,Zone,Number of Employees,Age of Business,Industry Group,Business Count,Employee Count,Marker
+24187818088,imports,year/2019,BE,B5,all,all,all,5170.0,4344200.0,
```
canwaf, line 2: Test a subset, not the entire fixture. Do we need this full CSV?

canwaf (changes requested):
We need to be more intentional with our fixtures, instead of using the files in whatever edge case work we need to make work. Can @rossbowen make synthetic data and metadata to focus the test case on this specific thing?

features/csvw.feature
```diff
@@ -51,6 +51,56 @@ Feature: Create CSVW metadata
...
+        <http://gss-data.org.uk/data/gss_data/trade/uk-trade-in-goods-statistics-by-business-characteristics#dimension/country>
+        <http://gss-data.org.uk/def/concept/cl-area/BE>;
+        <http://gss-data.org.uk/data/gss_data/trade/uk-trade-in-goods-statistics-by-business-characteristics#dimension/industry-group>
+        <http://gss-data.org.uk/data/gss_data/trade/uk-trade-in-goods-statistics-by-business-characteristics#concept/industry-group/all>;
+        <http://gss-data.org.uk/data/gss_data/trade/uk-trade-in-goods-statistics-by-business-characteristics#dimension/year>
+        <http://reference.data.gov.uk/id/year/2019>;
+        <http://gss-data.org.uk/data/gss_data/trade/uk-trade-in-goods-statistics-by-business-characteristics#dimension/zone>
+        <http://gss-data.org.uk/def/concept/cl-area/B5>;
+        <http://gss-data.org.uk/def/measure/trade> 24187818088;
+        <http://gss-data.org.uk/def/trade/property/attribute/business-count> "5170.0";
```
rossbowen, line 93: I'm noticing that numerics are being output as strings (which is the default behaviour - `csv2rdf` will treat things as strings unless told otherwise.

It's an extra ask, but I wonder if we should do something like this, but for attributes, so a user can explicitly state something is a number if they so wish. 

https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/csvw/mapping.py#L352-L353

Shannon95: Users can now explicitly state something is a number within the mapping if required.

rossbowen (approved):

PR #297, opened 2021-07-27
Issue 48/update pandas - bump pandas 🎉 

mikeAdamss:
Untangling the most tangley thing in the history of uber tangled tangly things.

* Switch to using tableset loaders from databaker (they work with modern versions of xlrd).
* Allows/have done pandas bump as _it_ requires that modern version of xlrd.
* Remove all spreadsheet using libs that require the _old_ version of xlrd.
* Fixed a few scrapers (we were getting `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` and `application/vnd.ms-excel` mixed in a few places, didn't used to matter, now it does).
* Removed ods conversion and monkey patch (handled by databaker natively now).
* Fixed a few tests
* Updated dependencies and setup.py, cleaned out a lot of libraries we don't need any more.

Have built databaker-docker with this gssutils in and ran some pipelines, all seems to be working as intended.

**NOTE - when we release we need to remove databaker from the databaker-docker pipfile, its coming through gssutils now, a conflict would be bad.**


ajtucker (approved):
Looks good!

My one concern is around the flip between passing around a stream vs. reading the whole stream into a BytesIO and passing that around. When we were dealing with large, presumably auto-generated Excel files, we ran into memory issues with this.

mikeAdamss:
cheers Alex, thats a really good point and one thats pretty hard to untangle. For now I'm gonna go ahead and merge but add a card for looking into mitigating memory issues (if we get an edge case in the meantime, we should be able to work around it in the transform easy enough).

PR #300, opened 2021-08-04
Add keywords to ONS scraper

rossbowen:


ajtucker (commented):
I think the logic for singleton vs list is accounted for elsewhere so you shouldn't need to change the function/lambda.

rossbowen:
Righto, will have a search! Before that it was returning `'"[\'item_1\', \'item_2\', \'item_3\']"@en'` so figured it wasn't accounting for it.

rossbowen:
So the tests make use of the `scraper.dataset.get_property()` method which I don't think is acting as expected with lists.

```python
from gssutils.scrape import Scraper
from gssutils.metadata import DCTERMS, DCAT, RDFS, namespaces

url = "https://www.ons.gov.uk/businessindustryandtrade/business/businessinnovation/datasets/foreigndirectinvestmentinvolvingukcompanies2013inwardtables"

scraper = Scraper(url)
ns = {'dct': DCTERMS, 'dcat': DCAT, 'rdfs': RDFS}.get("dcat")
scraper.dataset.get_property(ns["keyword"])
#> rdflib.term.Literal("['stocks', 'investment flows', 'business investment']", lang='en')
```

I think I'd expect the output to be a list of literals, not the list represented _as_ a literal, i.e., 

```python
[
    rdflib.term.Literal('stocks', , lang='en'),
    rdflib.term.Literal('investment flows', lang='en'),
    rdflib.term.Literal('business investment, lang='en')
]
```

Seems like the offending line is `L77`,

https://github.com/GSS-Cogs/gss-utils/blob/9fbc742ecd713e281d766a89681bc56b9a2211f6/gssutils/metadata/base.py#L72-L77

Because it runs the function specified in the metadata on the list as a whole, not on each element.

```python
prop, status, f = scraper.dataset._properties_metadata["keyword"]
f(scraper.dataset.__dict__["keyword"])
#> rdflib.term.Literal("['stocks', 'investment flows', 'business investment']", lang='en')
```

I'll probably change it to use `map` or something to iterate over the components of a list.

ajtucker (approved):
Thanks for figuring it out! Looks good.

Issue #238, opened 2021-04-20
ONS metadata comments/descriptions


ajtucker:
https://github.com/Swirrl/cogs-issues/issues/237 identifies some issues with comments and descriptions from the ONS `/data.json`.

mikeAdamss:
could you replicate the comments over here please Alex (don't have access to that)

ajtucker:
> could you replicate the comments over here please Alex (don't have access to that)

Sure, sorry. The issue is that https://beta.gss-data.org.uk/cube/explore?uri=http%3A%2F%2Fgss-data.org.uk%2Fdata%2Fgss_data%2Ftrade%2Fons-international-exports-of-services-from-subnational-areas-of-the-uk-catalog-entry shows an overly long description with markdown.

The [Epic](https://github.com/GSS-Cogs/gss-utils/issues/85) for this is to go through all the scrapers and make sure we're getting the `rdfs:comment` and `dct:description` fields right, along with using a special property for Markdown descriptions (see the metadata docs linked in the Epic).

The best bet here is to create an issue branch for this and add some BDD specifically for this dataset to check that the metadata we populate is right wrt. that revised metadata doc. That may break some of the other BDD tests, which we'll need to refactor and then may need to refactor the code in other scrapers!

rossbowen:
So this looks to be displaying correctly to me on `beta` but I think that's because of some massaging of the data which has gone on.

For the mentioned dataset, [this](https://github.com/GSS-Cogs/family-trade/blob/c2b42ac95922169dce57e91f88bea19b0839fefd/datasets/ONS-International-exports-of-services-from-subnational-areas-of-the-UK/main.py#L336-L355) looks to be the offending line, `comment` should be a short description which appears on the catalog page. `description` is not currently displayed anywhere.

In general, a quick review of the ONS scraper shows where stuff is populated from.

https://github.com/GSS-Cogs/gss-utils/blob/9fbc742ecd713e281d766a89681bc56b9a2211f6/gssutils/scrapers/ons.py#L51

https://github.com/GSS-Cogs/gss-utils/blob/9fbc742ecd713e281d766a89681bc56b9a2211f6/gssutils/scrapers/ons.py#L61-L64

Taking a typical dataset with `type = dataset_landing_page`, we get:

```python
import requests
url = "https://www.ons.gov.uk/economy/nationalaccounts/balanceofpayments/datasets/uktradeallcountriesseasonallyadjusted/data"
[...]

PR #301, opened 2021-08-05
Add better descriptions when scraping ONS ad-hoc releases

rossbowen:
Relates to discussion at https://github.com/GSS-Cogs/gss-utils/issues/238. Gives ONS ad-hoc releases shorter comments and completes the description (previously these were left blank).

Had some trouble with the `behave` test. The scraped description has `\n` newlines in it which behave escapes resulting in a failure. I've fudged it for now.


ajtucker (approved):
Seem like a good fudge!

Issue #85, opened 2020-09-29
rdfs:comment and dct:description


ajtucker:
The [metadata documentation for PMD](https://github.com/Swirrl/cogs-issues/wiki/PMD4-metadata/_compare/8af17d9d31c9d32176b4180e211b34a558c2b494...3465931d4b1f01a3b1e1f714acaefd051a53a478) has been updated to make clear how `rdfs:comment` and `dct:description` are used in the user interface.

The scrapers need to be updated with this in mind.

rossbowen:
[UK Sea Fisheries Statistics](https://staging.gss-data.org.uk/resource?uri=http%3A%2F%2Fgss-data.org.uk%2Fdata%2Fgss_data%2Fcovid-19%2Fmmo-ad-hoc-statistical-release-uk-sea-fisheries-statistics-catalog-entry) has:

```ttl
<http://gss-data.org.uk/data/gss_data/covid-19/mmo-ad-hoc-statistical-release-uk-sea-fisheries-statistics-catalog-entry>
    a <http://publishmydata.com/pmdcat#Dataset>;
    dcterms:description """Published in response to the COVID-19 pandemic to provide timely evidence on impacts on commercial sea fishing activity"""@en;
    rdfs:comment "Volume landed of the UK fishing fleet by country, vessel length and species group, November"@en;
```

I don't see `dcterms:description` used anywhere on PMD. `rdfs:comment` is used within the catalog and cube explore pages. The wiki talks about `pmdcat:markdownDescription` being available.
> ```pmdcat:markdownDescription # a longer description of the dataset for the About tab. Can contain markdown. If so, should have a datatype of <https://www.w3.org/ns/iana/media-types/text/markdown#Resource>.```

Having a quick look at https://www.gov.uk/government/statistics/final-uk-greenhouse-gas-emissions-national-statistics-1990-to-2019:

```python
#%%
import requests

x = (requests.get(
        f'https://www.gov.uk/api/content/government/statistics/final-uk-greenhouse-gas-emissions-national-statistics-1990-to-2019'
        ).json()
)
```

Yields:

```python
#%%
x["description"]

#> 'Final estimates of UK territorial greenhouse gas emissions.'

#%%
x["details"]["body"]

[...]

PR #302, opened 2021-08-06
Produce a pmdcat:markdownDescription triple for scraper descriptions

rossbowen:
Partially addresses https://github.com/GSS-Cogs/gss-utils/issues/85. While we discuss whether `dcterms:description` can be displayed on the about catalog, I've made it so the description is duplicated in a `pmdcat:markdownDescription` triple.

Will make descriptions start appearing on the about pages of cubes on PMD.


ajtucker (commented):
I'd been doing these kinds of overrides when setting attributes, e.g. https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/metadata/pmdcat.py#L62

ajtucker (approved):
Looks grand!

There are other places that it makes sense to pull out markdown or HTML and turn it into markdown, so we should raise another issue to check these other places use the same approach.

We need to have a think, from the point of view of a publisher, whether just using the markdown datatype on a `dct:description` is good enough and whether the publisher needs to know about specialized properties specific to PMD.

Issue #172, opened 2021-02-03
Feature: Geo/Time coverage/resolution metadata


rossbowen:
Supports: End users finding data

Some datasets include spatio-temporal coverage and resolution as metadata in the spreadsheet or methodology, while others will include this in the title of the dataset, e.g. "LSOA Electricity Consumption". Often this information is implicit in the context of the dataset.

As BAs collect the landscape and details of a dataset and then DMs manage the metadata and reference data details, we need to make the spatio-temporal coverage and resolution explicit such that it can be represented as dataset metadata using DCAT and extensions.

These properties could be added to Airtable, though the main place to record them should be the `info.json` file associated with the dataset.

- [ ] Ensure gss-utils metadata/scraper can pick up coverage/resolution and add it to the dataset metadata. GSS-Cogs/gss-utils#78
- [ ] Add a task to each dataset/transform issue to remind us to record this info. GSS-Cogs/airtable-utils#9

ajtucker:
See also https://github.com/Swirrl/cogs-issues/issues/92

ajtucker:
- [ ] Implement custom spatio/temporal metadata properties in gss-utils as per https://github.com/Swirrl/cogs-issues/issues/92
- [ ] Pick an example dataset and add values to those properties in the transform itself.

rossbowen:
https://github.com/GSS-Cogs/gss-utils/blob/a85ce1428cad003c5f9e05d0ecf821275ef4ddb3/gssutils/metadata/pmdcat.py#L43-L55

Additional properties can go here, and can be set as attributes against the scraper (`scraper.dataset.whatever`). We can provide some helper methods.

https://github.com/GSS-Cogs/gss-utils/blob/a85ce1428cad003c5f9e05d0ecf821275ef4ddb3/gssutils/scrape.py#L362-L363

https://github.com/GSS-Cogs/gss-utils/blob/a85ce1428cad003c5f9e05d0ecf821275ef4ddb3/gssutils/scrape.py#L401-L403

Robsteranium:
Is the proposal to add these as hard-coded fields in the `info.json`? Ultimately we ought to be able to derive these from the observation data which would mean they didn't need to be maintained by hand. I've provided some [example queries](https://github.com/Swirrl/cogs-issues/issues/92#issuecomment-861559688) which demonstrate how this might work.

canwaf:
[...]

PR #304, opened 2021-08-09
Add pmd-properties for spatio-temporal info 

rossbowen:
Related to https://github.com/GSS-Cogs/gss-utils/issues/172. Allows capture of bespoke spatio-temporal info (as `dcat` doesn't quite fit needs).


ajtucker:
I'm not sure I follow why we can't just say `dct:temporal` `<http://reference.data.gov.uk/id/gregorian-interval/2013-01-01T00:00:00/P7Y>`  ?

[DCATv2 says](https://www.w3.org/TR/vocab-dcat-2/#Property:dataset_temporal)
> The temporal coverage of a dataset may be encoded as an instance of dct:PeriodOfTime, or may be indicated using a URI reference (link) to a resource describing a time period or interval.

The reference.data.gov.uk resources are representations of time periods / intervals describing the start and end instants.

rossbowen:
Thought the same as I was working through it. Will check the reason again when Robin's back.

ajtucker:
`gssc:temporalStart` and end are essentially short-cuts to `dcat:temporal` / `svv:min` and max.

We could define them in these terms, although IIRC OWL2 doesn't allow mixed object & datatype property chains.

ajtucker:
As I understand @Robsteranium's concerns, I think it's about not wanting to special case the metadata display components too much? I.e. SPARQLing for `eg:dataset a qb:DataSet; ?p ?o` to get the metadata (modulo fetching labels).

I'd like to think that just using `dct:temporal` (note, not `dcat:` as above) and pointing to `reference.data.org.uk/id` resources would achieve the same thing, albeit with some potentially clunky labels for arbitrary time periods where ideally you'd want e.g. `2013 -- 2019`, but that's another issue.

In any case, I think the right thing to do here is to assert both `dct:temporal` and the two bespoke `gssc:temporalStart` and gssc:temporalEnd` properties in the metadata and compute one from the other(s).

rossbowen:
👍  Half way through a commit which will take `dcat:temporal` and derive the bespoke triples. Feels similar to the approach I put in place for `markdownDescription` last week doing it that way.

PR #305, opened 2021-08-10
Add better descriptions when scraping GOVUK releases

rossbowen:
Forms a longer markdown description from the details section of GOVUK publications (see [here](https://www.gov.uk/government/statistics/final-uk-greenhouse-gas-emissions-national-statistics-1990-to-2019) as an example).

One test failing currently because I can't get behave strings to behave how I would like - I think manually performing the test suggests it should work.


rossbowen:
`dcterms:description` isn't treated as a markdown literal atm.

https://github.com/GSS-Cogs/gss-utils/blob/f61322418b700917172b9a77ea2b9211d4c738cf/gssutils/metadata/dcat.py#L18

...but the `pmdcat:markdownDescription` is. I can open another issue to think about whether `dcterms:description` should be - it could affect other scrapers.

ajtucker (approved):
Have added some TODOs and pointers to a placeholder issue, to remind us.

Otherwise, let's do this thing :)

PR #306, opened 2021-08-10
Add qb:order to dimension components based on CSV column order.

ajtucker:


features/csvw.feature
```diff
@@ -636,4 +636,58 @@ Scenario: Manually Overriding CSV-W's graph URI works.
...
+    @prefix eg-comp: <http://gss-data.org.uk/data/gss_data/something/something#component/> .
+    @prefix sdmx-dimension: <http://purl.org/linked-data/sdmx/2009/dimension#> .
+    @prefix sdmx-attribute: <http://purl.org/linked-data/sdmx/2009/attribute#> .
+    @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+    # From https://www.w3.org/TR/vocab-data-cube/#attachment-example with a little re-arranging
+    eg:structure a qb:DataStructureDefinition;
+        qb:component eg-comp:place, eg-comp:year, eg-comp:sex, eg-comp:life-expectancy, eg-comp:unit .
+
+    eg-comp:place qb:dimension sdmx-dimension:refArea; qb:order "1"^^xsd:long .
```
rossbowen, line 687: `xsd:long`?

ajtucker: It's an oddness with `csv2rdf` that I'm going to raise a separate ticket for, as I think it should come through as just a simple `int` so we can say `qb:order 1` as per the example in the QB specs.

In any case, I guess it shouldn't matter to PMD as long as they're comparable and ordered / orderable.

rossbowen (approved):

features/csvw.feature
RickMoynihan: https://github.com/Swirrl/csv2rdf/issues/63

PR #315, opened 2021-09-03
Defra scraper

mikeAdamss:
Scraper +test for defras indicator datasets. Example: https://oifdata.defra.gov.uk/2.

After a chat with Alex and Ross we are treating each of these as a catalog page with each individual indicator page being linked to as dataset of a single distribution.

Best fit as its a bit of an odd one but you end up with a nice clean usable thing.

Have gone with the assumption that the page "1-2-3" has the csv "1-2-3.csv" as lack of this scraper is blocking things and (a) its a pretty safe assumption (b) I cant see a clever way to explicitly get that information out of the javascript tangle.

### DIFF
--- a/features/fixtures/cassettes/domains/oifdata.defra.gov.uk.yml
+++ b/features/fixtures/cassettes/domains/oifdata.defra.gov.uk.yml
@@ -1,0 +1,3108 @@
+interactions:
+- request:
+    body: null
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.26.0
+    method: GET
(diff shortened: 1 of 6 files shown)


ajtucker (commented):
Looks grand. I couldn't instantly see how to get the `.csv` from the page either, but reckon it's a safe assumption.

For the `append_to_host_url()` function, I've used `urljoin` elsewhere to do URI resolution as per https://datatracker.ietf.org/doc/html/rfc3986#section-5. The main thing is that we want to ensure we're doing the same thing that the user-agent (web browser) does, in that if you're on page `http://example.com/blah/de/blah.html` and have a link to `blah.csv`, then you end up going to `http://example.com/blah/de/blah.csv`.

mikeAdamss:
cheers Alex, have switched over to urljoin

ajtucker (approved):
:+1: Nice.

PR #316, opened 2021-09-07
Remove rdflib 4.2.2 restriction

robons:
Using the [pyRDFa](https://github.com/RDFLib/pyrdfa3) library to perform RDFa -> `rdflib.Graph` conversion.

This implicitly upgrades rdflib to the latest current version, 6.0.0

N.B. I have the [black](https://pypi.org/project/black/) autoformatter installed which has altered the layout of a few of the files I made changes in.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -16,6 +16,14 @@
         ]
     },
     "default": {
+        "appnope": {
+            "hashes": [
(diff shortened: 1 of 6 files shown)


ajtucker (approved):
Thanks @robons , looks good!

It's slightly tricky to see the semantic diffs without the syntactic re-formatting. I wonder whether we should just do the syntactic re-formatting for the whole project in a separate PR (not now, but in future)?

PR #317, opened 2021-09-14
Integrating csv-qb with Scraper class

robons:
Relates to https://github.com/GSS-Cogs/csvwlib/issues/28

I've added a function to the `Scraper` class which pulls the metadata from the relevant models parts. I believe it safely copes with all possible values which could be set. 

It's easy to serialise the CatalogMetadata to JSON. The way I'm expecting users to do this is as follows:
```python
...
from pathlib import Path

scraper = Scraper("http://some-uri")

catalog_metadata = scraper.as_csvqb_catalog_metadata()
catalog_metadata.to_json_file(Path("./out/catalog-metadata.json"))
```


canwaf (approved):
This is going to merge into a branch, and so I'm happy to approve it; however it it was to make its way to main we need to version our imports from our own software.

robons:
> This is going to merge into a branch, and so I'm happy to approve it; however it it was to make its way to main we need to version our imports from our own software.

I presume you just mean fixing at a particular commit id? Or do you mean making sure we upload to pypi and install from there?

PR #325, opened 2021-09-28
info.json v1.1 documentation

robons:
A lengthy treatise on the use of the info.json v1.1 syntax.

Satisfies https://github.com/GSS-Cogs/csvwlib/issues/118

I think the pieces potentially missing in this are:

* Discussion of pathification of CSV column title -> column name (for use in the URI templates).
* Discussion on the distinction between dimensions/dimension values and attribute/attribute values. I'm hoping this will be clear enough from context for me to be able to omit it.


canwaf (approved):
Good work, clear. Ties together well. Should be reviewed again after we identify a documentation style, etc.

PR #337, opened 2021-11-03
add primitive retries

mikeAdamss:
Getting issues where (we seem, it's hard to tell) to be getting 200 responses from a json endpoint on the ONS website but with no content in them (so spawns JsonDecode errors). This appears to happen inconsistently.

Not sure why.
Not sure why I can only recreate it in a container.
Not sure why its inconsistent (its not the same urls that break each time).
Not sure if its just the ONS scraper (hopefully).

Have wanted retry logic for a while anyway, so this PR adds optional (env var to turn it on) retry logic for the ONS scraper. 100% not _the_ solution, but a thing worth having and gets any effected ONS pipelines back to working in the meantime.

Also did a bit of a clean up since I was in there anyway. Lots of red but its just removing excessive comments (this was an example script back in the day) and replacing all the separate http gets with a single function.

Usage, just set an env var to turn it on, i.e `os.environ["ONS_SCRAPER_RETRIES"] = "True"` at the top of the script or set it in your terminal.

### DIFF
--- a/gssutils/scrapers/ons.py
+++ b/gssutils/scrapers/ons.py
@@ -1,21 +1,18 @@
-from distutils.util import strtobool
-import logging
-import os
-
 from csv import DictReader
 from io import StringIO
+import logging
+import mimetypes
+import os
+from urllib.parse import urlparse, urlunparse
 
 import backoff
 from dateutil import tz
 from dateutil.parser import parse, isoparse
-
-from urllib.parse import urlparse, urlunparse
-
+from distutils.util import strtobool
 
 from gssutils.metadata.dcat import Distribution
 from gssutils.metadata.mimetype import Excel, ODS, CSV, ExcelOpenXML, CSDB
 
-import mimetypes
 
 # save ourselves some typing later
 ONS_PREFIX = "https://www.ons.gov.uk"
--- a/gssutils/scrapers/ons.py
+++ b/gssutils/scrapers/ons.py
@@ -225,8 +225,6 @@ def handler_dataset_landing_page(scraper, landing_page, tree):
         # iterate through the lot, we're aiming to create at least one distribution object for each
         for i, version_dict in enumerate(versions_dict_list):
 
-            version_dict = versions_dict_list[i]
-
             version_url = version_dict["url"]
             issued = version_dict["issued"]
 
--- a/gssutils/scrapers/ons.py
+++ b/gssutils/scrapers/ons.py
@@ -1,10 +1,10 @@
 import logging
-import json
 import os
 
 from csv import DictReader
 from io import StringIO
 
+import backoff
 from dateutil import tz
 from dateutil.parser import parse, isoparse
 from retrying import retry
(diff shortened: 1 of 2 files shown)


gssutils/scrapers/ons.py
```diff
@@ -18,6 +20,32 @@
...
+def get(url, scraper):
+    """ Given a url, return a dict"""
+    r = scraper.session.get(url)
+    if not r.ok:
+        raise ValueError(f"Aborting. Issue encountered while attempting to scrape '{url}'. Http code"
+                        f" returned was '{r.status_code}.")
+    return r.json()
+
+
+@backoff.on_exception(backoff.expo, Exception, max_time=30)
```
robons, line 39: It might be nice to be able to configure the number of retries. I wonder whether 30 is so high that our IP address gets blocked for too much botty action. Also, I've noticed the `exponential` backoff.... how long does it take to complete the 30 tries?

I'm a little worried that could take hours which would block the Jenkins server up a bit.

gssutils/scrapers/ons.py
```diff
@@ -18,6 +20,32 @@
 ONS_TOPICS_CSV = 'https://gss-cogs.github.io/ref_common/reference/codelists/ons-topics.csv'
 
 
+def get_dict_from_json_url(url, scraper):
+    """ Wrapper to Let the DE decide if they want retries or not via an env var"""
+    if bool(os.getenv("ONS_SCRAPER_RETRIES", None)):
```
robons, line 25: IIRC, this `bool` conversion doesn't do what you would expect it to, use [bool(distutils.util.strtobool(some_string))](https://stackoverflow.com/a/18472142) instead

gssutils/scrapers/ons.py
```diff
@@ -28,34 +56,16 @@ def scrape(scraper, tree):
...
-    # Acquire title and description from the page json
-    # literally just whatever's in {"description": {"title": <THIS> }}
     scraper.dataset.title = landing_page["description"]["title"].strip()
 
-    # Same with date, but use parse_as_local_date() which converts to the right time type
     scraper.dataset.issued = parse_as_local_date(landing_page["description"]["releaseDate"])
 
-    # each json page has a type, represented by the 'type' field in the json
-    # the most common one for datasets is dataset_landing_page
-    # if that's the page type we're looking at then the comment is in {"description": {"summary": <THIS> }}
```
robons, line 56: Bye bye comments.

robons (approved):

gssutils/scrapers/ons.py
mikeAdamss: yeah you're right, str==true thing. good shout 👍

gssutils/scrapers/ons.py
mikeAdamss: its time in seconds, so it'll exponentially backoff to a max of 30 seconds per http request.

agree with the config thought btw, not one for this pr but a central configurable http client/method/function get for all the scrapers (or something along those lines) is probably worth doing.

canwaf:
The backoff library isn't that primitive, good to see it @mikeAdamss!

mikeAdamss:
I thought better and switched to backoff later in the life of this, trust me, v1 was ropey af 🤣

PR #341, opened 2021-11-17
allow passing of dicts/pd.DataFrame to build wrapper

mikeAdamss:
pr into a pr.

the assumption of using build() on local files makes it painful to integrate with other software (the sdg build tool) so added a utility to pass in dicts and/or a pd.DataFrame to build().

It's using the `tempfile` library so cleans up after itself (if you haven't used, the "tempfile" gets deleted when the pointer is garbage collected - tis nifty).

Shouldn't impact any of your other logic but should let me call `build()` in a python script without littering.

If you run the `./features/utils.feature` on it's own it passes.


mikeAdamss:
missed something

canwaf:
Did you needs this @mikeAdamss, or are we waiting on a new PR?

mikeAdamss:
nah, Rob said on the away day I can do what I need without the wrapper, I'll bend his ear when I'm back in.

PR #342, opened 2021-12-14
Muazzam chaud/issue211

MuazzamChaud:


gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -115,7 +115,7 @@ def _metadata_from_dict(config: dict) -> "CatalogMetadata":
         dataset_issued=dt_issued,
         theme_uris=theme_uris,
         keywords=config.get("keywords", []),
-        landing_page_uri=config.get("landingPage"),
+        landing_page_uris=config.get("landingPage"),
```
robons, line 118: The landing page can be either a string OR a list of strings here. If it's just a string we need to put it in a list.

unittests/configloaders/test_infojsontests.py
```diff
@@ -172,7 +172,7 @@ def test_cube_metadata_extracted_from_info_json():
     # landingpage - pass
 
     expected_landingpage = "https://www.gov.uk/government/statistics/bottles-bulletin"
-    actual_landingpage = cube.metadata.landing_page_uri
+    actual_landingpage = cube.metadata.landing_page_uris
```
robons, line 175: You're saying something of type `str` is equal to something of type `list[str]` here.

PR #344, opened 2021-12-16
Bump databaker

mikeAdamss:
bump databaker to `gss_speedup` commit `2dc3f373910a657aa506ba96ba0f2f0bdf488522 ` (with pyhamcrest pinned at <=2.0.2).

Pin CacheContol at 0.12.6 as later versions break the api scraper tests.

Update the makefile to get around a permissions error.

Note - have build databaker-docker on this gssutils commit and all the correct things are pulled in.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -16,14 +16,6 @@
         ]
     },
     "default": {
-        "appnope": {
-            "hashes": [
-                "sha256:93aa393e9d6c54c5cd570ccadd8edad61ea0c4b9ea7a01409020c9aa019eb442",
-                "sha256:dd83cd4b5b460958838f6eb3000c660b1f9caf2a5b1de4264e941512f603258a"
-            ],
-            "markers": "sys_platform == 'darwin'",
-            "version": "==0.1.2"
-        },
         "argparse": {
             "hashes": [
                 "sha256:62b089a55be1d8949cd2bc7e0df0bddb9e028faefc8c32038cc84862aefdd6e4",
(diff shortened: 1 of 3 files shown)


Pipfile.lock
```diff
@@ -1009,100 +1227,137 @@
...
             "hashes": [
-                "sha256:c203ec8783bf771a155b207279b9bccb8dea02d8f0c9e5f8ead507bc3246ecc1",
-                "sha256:ef9d7589ef3c200abe66653d3f1ab1033c3c419ae9b9bdb1240a85b024efc88b"
+                "sha256:04ff808a5b90911829c55c4e26f75fa5ca8a2f5f36aa3a51f68e27033341d3e4",
+                "sha256:d9bdec0013ef1eb5a84ab39a3b3868911598afa494f5faa038647101504e2b81"
             ],
-            "markers": "python_version >= '2.6' and python_version not in '3.0, 3.1, 3.2, 3.3'",
-            "version": "==2.4.7"
+            "markers": "python_version >= '3.6'",
+            "version": "==3.0.6"
```
robons: I believe this causes rdflib some problems. I've been pinning pyparsing at 2.4.7 in most places to get around that fact. It's possible that now we're using rdflib >=6.0.0 it isn't a problem anymore, but that's one to watch out for.

robons (approved):

setup.py
```diff
@@ -17,9 +17,9 @@
     install_requires=[
         "requests",
         "python_dateutil",
-        "CacheControl",
+        "CacheControl==0.12.6",
```
robons, line 20: I would be tempted to put an explaination in a comment against this so we know why we're doing it.

ajtucker (approved):
I had some weirdness with the Makefile when using `:/app` rather than `:app`, so good to see that you got this working.

All looks fine and it works for me and on Jenkins, so I guess all the tests still pass and it's good to go.

Why the pin on CacheControl btw? I saw one fix that [ensure we reset the file handle after reading](https://github.com/ionrock/cachecontrol/commit/d085f509811a9dcd83c416897553883b4fe7c29e) which reminded me of one of the issues you were having somewhere?

setup.py
mikeAdamss: good shout, will add something

mikeAdamss:
@ajtucker - I'm not sure how or why but later version of CacheControl are causing a write to closed file error for the API scraper tests.

Will need to figure that out but even getting that far (helping with Andrews spike) was a total pig, error are here if you want a nose: https://ci.floop.org.uk/job/GSS_data/job/gss-utils-branches/job/canwaf%252Frdflib-upgrade-spike/6/cucumber-html-reports/overview-features.html

mikeAdamss:
actually, given Robs mitigated this anyway I'll hold off on releasing until Monday, just in case we don't end up needing the pin.

PR #347, opened 2022-01-28
Include Validation Error Short URLs in Console Outputs#256

MuazzamChaud:
we are now using csvcubed-0.1.0-rc4 from pypi


Pipfile.lock
```diff
@@ -1183,7 +1165,7 @@
         },
         "csvcubed-devtools": {
             "git": "https://github.com/GSS-Cogs/csvwlib.git",
-            "ref": "3de392c500dfd9638efc8ed51ba8818193eb4e51",
+            "ref": "9cee2eaa73c62c4762108ead2b96bd4d7417ab81",
```
robons, line 1168: We can (and probably should) make sure csvcubed-devtools is installed from pypi too. I believe it's defined in the `Pipfile`.

gssutils/csvcubedintegration/infojson2csvqb/build.py
```diff
@@ -46,9 +47,9 @@ def build(
 
     if len(validation_errors) > 0 or len(json_schema_validation_errors) > 0:
         for error in validation_errors:
-            print(
-                f"{Fore.RED + Style.BRIGHT}Validation Error: {Style.NORMAL + error.message}"
-            )
+            print(f"{Fore.RED + Style.BRIGHT}Validation Error: {Style.NORMAL + error.message}")
+            if isinstance(error, SpecificValidationError):
+                print(f"Validation__error_URL: {error.get_error_url()}")
```
robons, line 52: Can we make sure to change this to something like `More information: `

setup.py
```diff
@@ -34,6 +34,7 @@
         "backoff",
         "vcrpy",
         "pyRdfa3",
+        "csvcubed-devtools==0.1.0rc4",
```
robons, line 37: This shouldn't be in here since it's a dev dependency - the setup.py file is only for dependencies say that users of this project need to install. The devtools stuff is only for our internal testing so it needs to remain in its location inside the Pipfile.

robons (approved):

PR #354, opened 2022-04-06
Adds a basic developer getting started

goofballLogic:


ajtucker (approved):
Looks good, thanks @goofballLogic .

For completeness, I wonder whether it's worth a section for non-devs? We'd been running all these different tools using pre-built Docker images.

goofballLogic:
> Looks good, thanks @goofballLogic .
> 
> For completeness, I wonder whether it's worth a section for non-devs? We'd been running all these different tools using pre-built Docker images.

Yes I must admit I thought of that, but as I have no idea how they use it I didn't know how. Who would be best to add that @ajtucker ?

PR #355, opened 2022-04-08
Scraper not brining in distributions as expected

santhosh-thangavel:
1. Bug fix. Using new class name for gov.uk scraper

#363


santhosh-thangavel:
@ajtucker Please review my pull request. Thank you.

ajtucker (approved):
Thanks @santhosh-thangavel, this looks to work ok.

I added a specific test case for the missing distributions for the HMRC RTS interactive analysis.

Next time around, we should create the (failing) test case first and then fix it :)

PR #366, opened 2022-08-12
Csvcubed datatypes

mikeAdamss:
Integrates the datatype setting functionality we're adding to csvcubed into our gssutils tooling.


mikeAdamss:
@robons - up for a PR when you get time, obviously wont merge anything until the csvcubed bits are in and all tests pass.

csvcubedintegrationfeatures/steps/infojson.py
```diff
@@ -14,23 +11,20 @@
...
-    context.json_schema_errors = json_schema_errors
+    context.validation_errors = validation_errors
+    context.json_schema_validation_errors = json_schema_validation_errors
 
 
 @Step("the CSVqb should pass all validations")
 def step_impl(context):
-    cube: QbCube = context.cube
-    errors = cube.validate()
-    errors += validate_qb_component_constraints(context.cube)
```
robons, line 32: I believe we still need to call both of these functions you've removed. They should get added to any `validation_errors` from the previous step definition.

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -14,42 +14,39 @@
...
     """
     Generates a QbCube structure from an info.json input.
     :return: tuple of cube and json schema errors (if any)
     """
-    with open(info_json, "r") as f:
+    
+    with open(info_json_path, "r") as f:
         config = json.load(f)
 
+    dtype = datatypes.get_pandas_datatypes(data_path, config=config)
```
robons, line 44: This function shouldn't work with the info.json v1.1 syntax files. It applies the logic of the qube-config.json and we need a similar function defined for info.json v1.1 files.

csvcubedintegrationfeatures/steps/infojson.py
```diff
@@ -14,23 +13,23 @@
...
     context.cube = cube_value
-    context.json_schema_errors = json_schema_errors
+    context.validation_errors = validation_errors
+    context.json_schema_validation_errors = json_schema_validation_errors
 
 
 @Step("the CSVqb should pass all validations")
 def step_impl(context):
     cube: QbCube = context.cube
     errors = cube.validate()
```
robons, line 30: These errors need to be concatenated with the others too.

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -131,44 +129,71 @@ def _metadata_from_dict(config: dict) -> "CatalogMetadata":
 
 
 def _columns_from_info_json(
-    column_mappings: Dict[str, Any], data: pd.DataFrame, info_json_parent_dir: Path
-) -> List[CsvColumn]:
+    column_mappings: Dict[str, Any], data_path: Path, info_json_parent_dir: Path
+) -> Tuple[Dict[str, str], List[CsvColumn]]:
```
robons, line 133: I think your types are the wrong way around in here.

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -131,44 +129,71 @@ def _metadata_from_dict(config: dict) -> "CatalogMetadata":
...
     defined_columns: List[CsvColumn] = []
+    dtype_mapping: Dict[str, str] = {}
+
+    data = pd.read_csv(data_path)
```
robons, line 137: Do we need to read the whole CSV. Can we limit it to 1 or 2 lines just to get the headers?

Dockerfile
```diff
@@ -27,6 +27,14 @@ RUN poetry export --format requirements.txt --output /requirements.txt --without
 RUN ${VENV_PIP} install --requirement /requirements.txt
 RUN ${VENV_PIP} install poetry
 
+# The install from requirements.txt is cirumventing the pin to cacheControl 0.12.6 as it lacks
+# sufficiant dependency resolution (something somewhere is bumping us to 0.12.11).
+# Unfortunetly when the "poetry install" in the ENTRYPOINT tries to resolve this on Jenkins it errors
+# while attempting the downgrade of 0.12.11 -> 0.12.6.
+# As a work around we're directly uninstalling cachecontrol (0.12.11) here with the venv
+# pip - poetry in the ENTRYPOINT will then install the required version.
+RUN ${VENV_PIP} uninstall cachecontrol -y
```
robons, line 36: It might be worth removing `--without-hashes` on line 24 to see whether that gives the requirements.txt enough of a constraint to install the appropriate version. I can't remember exactly why I included `--without-hashes` there; I remember the build was failing with it, but I have a suspicion that it was because we were installing packages directly from github repos (which we aren't anymore). May be worth a try.

robons (approved):
Just one or two minor things to consider, but seems to be pretty much there. Good work.

Dockerfile
mikeAdamss: Worth a go but no can do, we've still got databaker from github in the mix.

Our version of databaker is such a gigantic bag of hacks right now that I don't see any way of merging that to push up to the official databaker pypi any time soon (its got packages in there that were depreciated in 2016, it works for us, but mainly as I've twisted it to our specific use case of new pandas and new python).

Basically, needs replacing and we don't have 6 months to spend on a tool for flattening spreadsheets we ultimately don't want..... just gotta hope it doesn't implode at some point before we stop using it I think.

gssutils/csvcubedintegration/configloaders/infojson.py
mikeAdamss: @robons - great question - this ended up opening a whole (necessary) barrel of worms on me - TLDR is I was (a) reading in the csv to get the QbColumns and then (b) using the data types to reading in the obs data.

but - that first read for QbColumns was also creating the codelists (I didn't realise originally), so allowed a scenario of pandas "best guess" typed codelists with correctly typed obs columns.

had to pull it apart hard to get the behaviour we want, tests pass but wanted to wait on a sanity check/re-review before I release as I had to change a lot.

would probably benefit from maybe a class and not having to repeat setting the `maybe_x` variables, but its internal glue code and I'm a fair bit overdue looking at other things so didn't go down that road.

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -9,47 +9,42 @@
 from pathlib import Path
 import json
 import copy
+from unittest.mock import NonCallableMagicMock
```
robons, line 12: Do we need this here?

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -86,16 +82,89 @@ def _override_config_for_cube_id(config: dict, cube_id: str) -> Optional[dict]:
...
-    metadata = _metadata_from_dict(d)
-    transform_section = d.get("transform", {})
-    columns = _columns_from_info_json(
-        transform_section.get("columns", []), data, info_json_parent_dir
-    )
+    """
+    There is some convoluted but important ordering to be considered here,
+    we need to:
+
+    1.) Get the column titles only from the csv
```
robons, line 91: I don't think we actually need the column titles at all. We're not doing conventional column configuration in gss-utils.

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -86,16 +82,89 @@ def _override_config_for_cube_id(config: dict, cube_id: str) -> Optional[dict]:
...
+
+    all_columns = column_titles_in_data + columns_titles_not_in_data
+
+    # step 2
+    column_schemas = {}
+    for column_title in all_columns:
+        maybe_config = column_mappings.get(column_title)
+        column_schemas[column_title] = _get_column_schema(column_title, maybe_config)
+
+    # raise Exception(column_schemas)
```
robons, line 129: remove

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -130,45 +199,120 @@ def _metadata_from_dict(config: dict) -> "CatalogMetadata":
...
-        maybe_config = column_mappings.get(col_title)
-        defined_columns.append(
-            _get_column_for_metadata_config(
-                col_title, maybe_config, data[col_title], info_json_parent_dir
-            )
-        )
+def _get_column_schema(
+    column_name: str,
+    col_config: Optional[Union[dict, bool]],
+) -> schema.SchemaBaseClass:
```
robons, line 205: I think the valid type signature is `Type[schema.SchemaBaseClass]` since you're passing around a type rather than in instance of `schema.SchemaBaseClass`.

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -86,16 +82,89 @@ def _override_config_for_cube_id(config: dict, cube_id: str) -> Optional[dict]:
...
+
+    Return a mapping of csvw spec datatypes to pandas primitive datatypes.
+    The mapping is provided by ACCEPTED_DATATYPE_MAPPING from csvcubed itself.
+    """
+
+    dtypes = {}
+    for column_title, column_schema in column_schemas.items():
+        
+        if hasattr(column_schema, "data_type"):
+            dtype_str = column_schema["data_type"]
```
robons, line 156: Given that your're actually holding a `Dict[str, Type[schema.SchemaBaseClass]]`, the column schema is never actually an instance of `schema.SchemaBaseClass` and so you'll never be able to get the `data_type` field off it.

gssutils/csvcubedintegration/configloaders/infojson1point1/mapcolumntocomponent.py
```diff
@@ -77,10 +76,10 @@ def map_column_to_qb_component(
...
     else:
-        raise ValueError(f"Unmatched schema model type {type(schema_mapping)}")
+        raise ValueError(f"Unmatched schema model type {schema_mapping}")
 
 
-def _from_column_dict_to_schema_model(
+def from_column_dict_to_schema_model(
     column_title: str,
     column: dict,
 ) -> Union[
```
robons, line 85: Again, this is really `Union[Type[schema.NewDimension], Type[...], ...]`.

robons (changes requested):
I don't understand how you behaviour test is passing, but I don't think you're actually loading the attribute/obsval columns' data types.

gssutils/csvcubedintegration/configloaders/infojson.py
mikeAdamss: sorry, didn't push my tidy up commit, that was silly.

mikeAdamss:
Good spots, should be better now, have also tidied up the logic (a bit) to make it a little easier to follow. Updated the dependencies as dependabot is moaning.

Unexpected BDD passing was because I was using ideal data (pandas best guess happened to return the correct results), should have started from a failed test. Have altered the test data dims to have leading zeroes, should blow up now if the the datatype functionality fails.

mikeAdamss:
Test was passing as I gave it perfect test data (so pandas "best guess" was right). Have updated to include some nice leading 0 dimensions.

Addressed pr comments, tidy up logic (a bit) and updated deps.

gssutils/csvcubedintegration/configloaders/infojson1point1/columnschema.py
```diff
@@ -419,3 +423,114 @@ def _get_new_attribute_values(
...
     return []
+
+
+@dataclass
+class InfoForColumn:
+    """
+    Convenience class to make the tracking and construction
+    of QbColumn objects easier to work with and understand.
+    """
+    name: str
```
robons, line 434: I'd call this `title` for consistency.

gssutils/csvcubedintegration/configloaders/infojson1point1/columnschema.py
```diff
@@ -419,3 +423,114 @@ def _get_new_attribute_values(
...
+    config: dict
+    schema_type: Optional[Union[Type[SchemaBaseClass], Dict]] = None
```
robons, line 436: This can also be expressed as `Union[Type[SchemaBaseClass], Dict, None]`.

gssutils/csvcubedintegration/configloaders/infojson1point1/columnschema.py
```diff
@@ -419,3 +423,114 @@ def _get_new_attribute_values(
...
+        return self.config.get("datatype")
+    
+    def instantiate_schema(self) -> SchemaBaseClass:
+        """
+        Instantiate and return the known schema with
+        the known mapping.
+        """
+        return self.schema_type.from_dict(self.config)
+
+    def with_column_schema(self) -> InfoForColumn:
```
robons, line 481: I'd be tempted to put set `schema_type: Union[...] = field(init=False)` at the top and put this logic in a `__post_init__` function which will automatically get called after init - you can then populate the schema_type field directly. Unless the copy functionality is truly necessary?

gssutils/csvcubedintegration/configloaders/infojson.py
```diff
@@ -86,16 +81,92 @@ def _override_config_for_cube_id(config: dict, cube_id: str) -> Optional[dict]:
...
+def _get_dtypes_from_schemas(info_for_columns: List[schema.InfoForColumn]) -> Dict[str, str]:
+    """
+    Return a mapping of csvw spec datatypes to pandas primitive datatypes.
+    The mapping is provided by ACCEPTED_DATATYPE_MAPPING from csvcubed itself.
+    """
+
+    dtypes = {}
+    for column_info in info_for_columns:
+        
+        if column_info.schema_type is schema.ObservationValue:
```
robons, line 154: The data_type should default to `decimal` for both ObservationValue columns as well as AttributeLiteral columns too.

robons (approved):
I sincerely hope I never have to look at this code again.

gssutils/csvcubedintegration/configloaders/infojson.py
mikeAdamss: As far as I can tell we've only ever create literals when `literalValuesDataType` is not None. As such have added some logic to catch that data type field but haven't added a default, I don't think it'd ever play (if it ain't specified, it ain't a literal, in gssutils land anyway).

There's a case for pulling apart all this implict stuff like we did in csvcubed configloader but (a) its glue code and (b) I'm _waaaaaaaaay_ overdue on my urgent switch of focus, if it needs doing it can't be me.

PR #367, opened 2022-11-01
documetnation for filter one

mikeAdamss:
some code comments to give some sort of indication on what filter_one does.


mikeAdamss:
not necessary

PR #390, opened 2022-12-20
Muazzam chaud/issue527

MuazzamChaud:


.github/workflows/updateCsvcubedPackage.yaml
```diff
@@ -0,0 +1,94 @@
...
+        id: Checkout_files_from_master_branch
+        run: git clone 'https://csvcubed:${{ secrets.CSVCUBED_GIT_COMMIT_TOKEN }}@github.com/GSS-Cogs/gss-utils.git' --depth 1
+
+      - name: Fetching csvcubed version from pypi and pyproject.toml
+        id: Fetching_csvcubed_versions
+        run: | 
+          step_success=false
+          for attemptNumber in 0 1 2 3 4 5
+          do
+            sleep $((2**$attemptNumber))m
```
robons, line 26: Can you put a comment in explaining why you need to sleep and retry?

robons (approved):
I've got some questions and suggestions, but in general this approach seems sensible.

